PR #2833, opened 2014-12-24
Update metisMenu

jkrehm:
Also add update.json (there's a pull request that seems to have done this, but I don't see it in the repository).


pnommensen:
The change in letter case will cause issues for some users. @jimaek thoughts?

jkrehm:
I noticed that after I sent the pull request. The library's author has the
difference in case, which may make the auto update never work. I haven't
had time to alter the pull request. I'm actually not real sure what should
happen. It's sort of frustrating that the automated update can never work
because of the difference in case.

jimaek:
We can do that after a major version update. 1.1.2>1.1.3 cant introduce any breaking changes.
Maybe we can change the format when 1.2 comes out.

PR #2886, opened 2015-01-01
add jPlayer v2.9.2

lpinca:
Update jPlayer to version 2.9.2


lpinca:
> jPlayer Repository Refactor: All the circle player specific files are now in the lib/circle-player folder.

I didn't include that skin as it is no longer present in the default `dist` folder and now lives in a separate folder with its own js files.
It's probably better to ship it as a separate project.

jimaek:
Thank you

PR #2887, opened 2015-01-01
update.json for webshim

isaackwan:
Webshim update.json

The jsdelivr repo is a few versions behind actually. Am I expected to manually upload them or not?


megawac:
Looks good and only if you have a need for one of those missing versions, otherwise its all good. Thanks!

PR #2905, opened 2015-01-02
Created jquery.cssmap js project

Arg0n4th:
Created jquery.cssmap js Project


Arg0n4th:
The url https://gist.github.com/choyer/3278059 seems to be a github url, why the automatic check fails?

Thanks

jimaek:
Because its looking for a real github project URL and not a gist.

PR #2906, opened 2015-01-02
Commit to delete wrong commits

Arg0n4th:
Is this the right way to rollback?


jimaek:
No, you didnt commit anything to this repo, there is nothing to roll back.

PR #2907, opened 2015-01-02
Rate-it - jquery 5 stars ranking system

Arg0n4th:
Rate-it - jquery 5 stars ranking system


jimaek:
What is the source of these files? The official repo has a different structure

Arg0n4th:
mmh , they moved. When I took these files there was a github project active.

Now the actual version is different. I think these file may be useful for backward compatibility. If you think they violate the requirements I will delete them.

jimaek:
Where did you download these files?

Arg0n4th:
I don'r remember sincerely, but I guess on rate-it github repo when it was active.

jimaek:
The latest files are located over here http://rateit.codeplex.com/
So if you want you can use them instead. Or you can close the PR.

jimaek:
Again the formatting in that repo is different. There are no css/ js/ folders plus there are images included in the css. So either we host everything or we host only js

PR #2914, opened 2015-01-03
Add snabbt.js update.json

as-com:
This is the perfect way to engage a new user with jsDelivr - new and popular project, and now automatically hosted on a fast CDN. :smile:


jimaek:
Awesome, thanks

PR #2918, opened 2015-01-03
Added version 3.8.1 of the plugin fixing folder structure

Fede72:
Added version 3.8.1 of the plugin fixing folder structure


Fede72:
There are 2 requirements in conflict. The info.ini requires the folder js/jquery.fileupload.js, the 5.19.7 file structure requires the query/js/jquery.fileupload.js.

Which is the right requirement I have to meet?

jimaek:
Why are you adding such an old version? The latest is 9.9.0

PR #2945, opened 2015-01-06
Lytebox JS library content viewer

Fede72:
Lytebox is a lightweight, cross-browser compatible and mobile friendly
Javascript library and content viewer


Fede72:
The file are correct also with Unescaped unicode characters

PR #2950, opened 2015-01-06
Lytebox JS library content viewer

Fede72:
Lytebox is a lightweight, cross-browser compatible and mobile friendly
Javascript library and content viewer


Fede72:
The files are both correct, there is a regex with these chars

jimaek:
Why did you change the file format? The official download has a different structure

PR #2966, opened 2015-01-07
add clappr-rtmp plugin

flavioribeiro:


files/clappr-rtmp/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "clappr-rtmp-plugin",
+  "repo": "flavioribeiro/clappr-rtmp-plugin",
+  "files": {
+    "basePath": "dist/",
+    "include": ["rtmp.min.js", "rtmp.js", "./assets/**/*"]
```
jimaek: If you are going to include everything in `dist/` then you dont need to use the `include` parameter. `"basePath": "dist/"` is enough and will add everything that its in that folder

jimaek:
Can you rename the directory to `clappr.rtmp` please? And squash your commits

files/clappr-rtmp/update.json
flavioribeiro: the problem is that the plugin build generates not only these files but more on dist/. This include will filter only the necessary ones in case the others are committed/pushed.

flavioribeiro:
done!

files/clappr-rtmp/update.json
jimaek: I see, can you then remove `rtmp.js` as well? The min is enough for production I think

jimaek:
Thank you

PR #2998, opened 2015-01-08
Adding Unveil Effects library

travoltron:


jimaek:
Your info.ini is broken

PR #3050, opened 2015-01-13
Fixed font bugs

Artur-A:
1) Chrome Mobile disturbs fonts if font base size is small. Base size
was 16 points, now extended to more than 2048 points.

2) Some circled icons wrongly was moved to the left side. This affected
only Firefox which cuts off left part of this icons. Now fixed.


jimaek:
thanks

PR #3101, opened 2015-01-16
Add minified version of polymer

as-com:
There's a minfied version on cdnjs, so I thought I might as well steal it...


jimaek:
We have auto-updates enabled for this project, this file would break the structure because they dont have it in their repo

PR #3104, opened 2015-01-17
+ KeyCDN

tomByrer:


jimaek:
Thank you :) 
Btw are you interested in maintaining our wiki?
it would be great to have everything documented

PR #3105, opened 2015-01-17
new wavesufer.js

tomByrer:
I should really automate this someday...

### DIFF
--- a/files/wavesurfer.js/info.ini
+++ b/files/wavesurfer.js/info.ini
@@ -1,0 +1,5 @@
+author = "katspaugh"
+github = "https://github.com/katspaugh/wavesurfer.js"
+homepage = "http://www.wavesurfer.fm/"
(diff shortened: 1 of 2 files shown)


jimaek:
In this case ` wavesurfer-js-map.json` is useless because it tries to load files form /src which we dont host. So you should update update.json to load only wavesurfer.min.js

tomByrer:
> update update.json to load only wavesurfer.min.js

Browsers (at least Chrome) will [error](http://stackoverflow.com/questions/18365315/jquerys-jquery-1-10-2-min-map-is-triggering-a-404-not-found).  Let me contact dev first please.

jimaek:
Well it will error in any case since we dont host src/

tomByrer:
Sorry about the 3 commits; still using web client.

megawac:
@tomByrer why don't you just fork from the initial commit xD

https://github.com/megawac/jsdelivr

PR #3106, opened 2015-01-17
Add jquery.serialScroll

bdukes:


files/jquery.serialscroll/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "bower",
```
jimaek: if you are going to use bower then the `name` parameter is used to query bower for the package. `jquery.serialScroll` is not the correct name in that case. You can switch to github instead if you want

bdukes: I'm seeing it show up on http://bower.io/search/?q=jquery.serialScroll and it installs fine for me with `bower install jquery.serialScroll`.  Am I missing something?

jimaek: Strange, when I looked it has 0 results. Then its looks good :)

PR #3133, opened 2015-01-20
[author] update angular.textAngular to v1.3.0 format

SimeonC:
Change the include paths to include all the new paths in the v1.3.0 release.
Pre-releases aren't included so this should have no effect until I release the v1.3.0 tag.

### DIFF
--- a/files/angular.textangular/update.json
+++ b/files/angular.textangular/update.json
@@ -3,6 +3,9 @@
   "name": "textAngular",
   "repo": "fraywing/textAngular",
   "files": {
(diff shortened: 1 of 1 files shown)


jimaek:
Maybe in the new version we could only host dist/ ? What do you think?

SimeonC:
We could do, There's a required CSS stylesheet that's only in the src, so the dist/ and that file?

jimaek:
Yeah, lets do that. Can you update the file and squash the commits?

SimeonC:
Update and squashed.

PR #3140, opened 2015-01-21
update info.ini and add update.json for bxslider

Tidal-Wave:


files/bxslider/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "bxslider-4",
+  "repo": "stevenwanderski/bxslider-4",
+  "files": {
+    "include": ["jquery.bxslider.min.js", "jquery.bxslider.js", "./images/*", "./plugins/*"]
```
jimaek: I dont think that `jquery.bxslider.js` is needed, the min file is enough right?
Also what about the css file? Its not needed in production?

Tidal-Wave: Im sorry, I was recently put in charge of this repo and am moving very quickly to address a lot of issues because it has not been maintained so I apologize. The .js file was meant to be the css. I will fix and squash.

jimaek:
Awesome, thank you

PR #3150, opened 2015-01-22
AmaranJS

hakanersu:


jimaek:
Since its a jquery plugin can you rename the directory to `jquery.amaran` please?
Also the css files are not needed?

hakanersu:
Sure i will fix it.

PR #3180, opened 2015-01-24
1.4.3

pingram3541:


pingram3541:
New to this fork/pull stuff...just wanted to update the version.

jimaek:
1. You cant modify existing files
2. You need to squash your commits into 1

pingram3541:
* Looks like the info.ini has a few problems
   * Where is the jquery.layout's mainfile (jquery.layout.min.js) in v-1.3.0rc30.79
      * It never existed because it was renamed previously with -latest which is no longer correct.
* Might be nothing but I have some concerns about the files in your submission...
   * Is jquery.layout.min.js (1.4.3) properly minimized?
      * It is but preserving the author credentials as the previous version was allowed to do
* We try to keep the file structure as consistent as possible between versions; there were some changes in file structure from previous versions...
   * In jquery.layout between 1.3.0rc30.79 and 1.4.3 there was a change in file structure
      * Again, this was due to -latest being appended to the older version, nothing else changed

All I did was rename the older version to no longer have -latest in the file name and this was only due to the bot complaining it no longer matched with the updated ini file.  Your documentation says that you will change the files to -latest so I did not modify the new versions from the original authors file names.  This is proving to be a more difficult process and if others have this difficulty I wonder how many just give up and walk away from the whole thing.  This is a great service with a great idea but seriously, I had to fork/copy a HUGE library to my local storage just to update 3 files only a few k and even then the process is frustrating.  Maybe I'm just a noob and need better instruction.  What is a squash besides a vegetable?  Is that a git term like fork...what is all this food reference?

jimaek:
Sorry I was traveling and didnt had time to respond.
Most of the bot complains can be ignored and are there to help the work of moderators.

I understand that its not easy thats why you can always open an issue asking for someone else to update a project for you. 

It is also why we added the auto-update feature, all you have to do is add an update.json file using the web ui and our bot will upload the latest version automatically. No need to fork or download anything.

If you still want to finish this PR then you will need to squash your commits before I can review them. Here are the instructions https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

PR #3184, opened 2015-01-24
Add Riot

aurri:
Thanks.


aurri:
Fixed author field and squashed into a single commit.

files/riot/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "riotjs",
+  "repo": "muut/riotjs",
+  "files": {
+    "include": ["riot.js", "riot.min.js"]
```
jimaek: I dont see these files here https://github.com/muut/riotjs/tree/v2.0.1
The bot will fail to update them in the future

jimaek:
Also I think its safe to remove the unminified file. In production only the min is needed.

files/riot/update.json
aurri: Yea, these files weren't there in previous versions. They're already in `master` however, and will be included in all upcoming releases from now on.

aurri:
Ideally, I would love to keep the unminified file as it can provide better feedback during the development (e.g. more meaningful errors) than the minified one. But if that's a requirement, no problem at all.

Let me know.

Thanks!

jimaek:
Looks good then, thank you!

aurri:
Awesome, thank you!

PR #3208, opened 2015-01-27
Add m8tro-bootstrap

idleberg:


files/m8tro-bootstrap/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "m8tro-bootstrap",
+  "repo": "idleberg/m8tro-bootstrap",
+  "files": {
+    "include": ["./dist/m8tro.css", "./dist/m8tro.min.css"]
```
jimaek: Thats not correct, your settings will create a folder called "dist".
Do this instead:
```
"files": {
    "basePath": "dist/",
    "include": ["m8tro.min.css"]
...
```

Also do you really need the un-minified file? Maybe its best to only hos the min.css?

idleberg:
To be honest, I keep wondering why Bootstrap or Bootswatch put their unminified CSS on CDNs. I don't mind getting only the minified version hosted.

jimaek:
One more change, please rename the directory to `bootstrap.m8tro` after that squash your commits and its good to go.

Thanks

idleberg:
Any specific reason why I should rename this? It's using the same name on Bower and npm, so I think it makes sense to keep it consistent.

jimaek:
To follow the existing format
https://db.tt/mJ4KKaPW
https://db.tt/Q47Mgus5

{mainLib}.{pluginName}

jimaek:
Nice, please squash the commits into 1 and then I can merge

PR #3215, opened 2015-01-27
json2

peteyork:
Adding the latest json2.js from https://github.com/douglascrockford/JSON-js


jimaek:
Instead of using the data just create a `0.2` version.

peteyork:
All right, you're the boss.. the dir name is now changed.

PR #3220, opened 2015-01-27
Add Evil Icons

outpunk:
This PR adds Evil Icons http://evil-icons.io


jimaek:
Do you need the unminified files? Maybe its best to only host the production files?

outpunk:
Well, you're right.

jimaek:
Dont you also need the svg files to make it work? I am not 100% sure how this project works :)

outpunk:
Nope, we don't need the svg files — the sprite is built into the js.

jimaek:
Awesome then, please squash your commits into 1 and its good to go

jimaek:
Thank you

outpunk:
Thanks!

PR #3221, opened 2015-01-27
Add lodash 3.0.0

yasyf:
Note that only the `modern` and `compat` builds are offered now.


megawac:
Can you include some of [the other builds](https://lodash.com/custom-builds).

/cc @jdalton 

P.S. @jdalton you may want to update this wiki https://github.com/lodash/lodash/wiki/build-differences some time in the future
P.S.S. I see it in the changelog but why did you drop support for the BB build?

jimaek:
Damn, I miss the `dist/` dir, it would make auto-updating so easy.

megawac:
Yeah @jimaek we'll have to update the `libgrabber` examples

jimaek:
I dont think that we can auto-update now with npm because its 2 different packages.
https://www.npmjs.com/package/lodash-compat + https://www.npmjs.com/package/lodash

So we either do manual updates or wait for @jdalton to see if he wants to make these files available in github.

jdalton:
If you can map files then for the cdn it would be:

[lodash.js](https://raw.githubusercontent.com/lodash/lodash/3.0.0/lodash.js)
[lodash.min.js](https://raw.githubusercontent.com/lodash/lodash/3.0.0/lodash.min.js)

[lodash.compat.js](https://raw.githubusercontent.com/lodash/lodash-compat/3.0.0/lodash.js)
[lodash.compat.min.js](https://raw.githubusercontent.com/lodash/lodash-compat/3.0.0/lodash.min.js)

jimaek:
I cant, 1 project can have 1 git/npm/bower source. If you want we could create a separate project called `lodash.compat` on jsDelivr but I dont really like the idea, it will be confusing for people.
The other alternative is doing manual updates.

What do you guys think?

jdalton:
Or create a script to do it :smile_cat:

jimaek:
Nah, a script wont work, we have a universal auto-updating solution https://github.com/jsdelivr/libgrabber

I guess for now manual updates will work just fine.

PR #3242, opened 2015-01-29
Changed Riot's update.json (added more files)

aurri:
The next Riot version will include new files: riot+compiler.js and riot+compiler.min.js.
Added them to update.json.

### DIFF
--- a/files/riot/update.json
+++ b/files/riot/update.json
@@ -3,6 +3,6 @@
   "name": "riotjs",
   "repo": "muut/riotjs",
   "files": {
-    "include": ["riot.js", "riot.min.js"]
+    "include": ["riot.js", "riot.min.js", "compiler.js", "compiler.min.js"]
   }
 }


jimaek:
Any chance to change the filenames? I am afraid that the + symbol could break stuff.

aurri:
Yea, it's a bit unconventional, but I'm not aware of any evidence against using `+` in urls at least. Or do you mean it can break stuff on jsDelivr? Maybe that's a good change to find out then? :)

jimaek:
It will definitely break the combination features when more than files are combined from a single project.
Example URL:
http://cdn.jsdelivr.net/g/jquery@2.1.0,angularjs@1.2.14(angular.min.js+angular-resource.min.js+angular-animate.min.js+angular-cookies.min.js+angular-route.min.js+angular-sanitize.min.js)
([docs](https://github.com/jsdelivr/jsdelivr#load-multiple-files-with-single-http-request))

Technically I can merge it but then this feature wont work for riot

aurri:
Well, you're right. Thanks for explaining!

Seems we could get the same result by just including `riot.min.js` and `compiler.min.js` and requesting them both with: http://cdn.jsdelivr.net/g/riot@2.0(riot.min.js+compiler.min.js)

Going to make that change then in a moment.

Thanks again.

aurri:
Updated the filenames.

jimaek:
Awesome, thank you!

PR #3247, opened 2015-01-29
Add waypoints 3.1.0

cappslock:
This PR adds waypoints 3.1.0.

This project also exists in jsdelivr as jquery-waypoints, but that project has been abandoned in favor of this project (see https://github.com/imakewebthings/jquery-waypoints).


files/waypoints/update.json
```diff
@@ -0,0 +1,22 @@
+{
+  "packageManager": "github",
+  "name": "waypoints",
+  "repo": "imakewebthings/waypoints",
+  "files": {
+  	"include": [
```
jimaek: All these files are in the lib/ directory. You need to use the "basePath" parameter to set it.

Also instead of listing all files you can simply do `"*.min.*"` and `./shortcuts/*.min.js`

PR #3254, opened 2015-01-29
adding socket.io 1.3.2

keithblaha:
Thanks for your continued efforts!


keithblaha:
Seems like it should be fine- please let me know if otherwise

PR #3270, opened 2015-02-01
adding hyphenator 4.3.0

piccaso:
Hyphenator 4.3.0 
Released on October 17, 2014 ( https://code.google.com/p/hyphenator/ )


piccaso:
These are the original 'patterns' from the release zip...
Looks like 4.2.0 (on jsdelivr) has them included the same way and as far as i can tell it works fine.
Am i supposed to escape these characters to get 4.3.0 included?

megawac:
Thanks @piccaso

PR #3275, opened 2015-02-01
Adding highlight.js loader v5.0

alidbg:
Added Highlight JS Loader Version 5.0


files/highlightjs.loader/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Ali.Dbg"
+github = "https://github.com/alidbg/highlight.js_loader"
+homepage = ""
```
megawac: Feel free to update this with the github link if you want

PR #3333, opened 2015-02-05
Adding 3 Libraries from AmCharts

abishekrsrikaanth:
Adding the following 3 libraries
- Amcharts
- AmMaps
- AmStockCharts

All the 3 of them are opensource charts from http://www.amcharts.com/.
They don't have a github repository so unable to setup auto update.


abishekrsrikaanth:
@jimaek Can you please advice on how to resolve the issues above.
1. There is a ini file on ammap, but the bot says it doesn't exist
2. Binary and unicode warnings.

jimaek:
I am not sure that we can host them because that would simplify the piracy of their script. They are actually selling their licenses and I dont want jsDelivr to have any problems.

abishekrsrikaanth:
@jimaek. the commited version is the free version that is available for download from http://www.amcharts.com/download/. This particular library is also on cdnjs and i wanted it to be available on jsdeliver as well.

jimaek:
Some mistakes that you made:
1. 1 PR per project
2. Host as little files as possible. We dont need 875 files in production
3. Always squash your commits

abishekrsrikaanth:
1. Not sure what you mean by 1 PR per project. Do you mean only 1 commit?
2. Most of the files are seperated by languages as the library supports i18n. Do you want me to concat all the files. Since the library is distributed as different files for convenience to select the different files individually could you advice on what I should do.
3. I can squash my existing commits into 1.

Please advice

jimaek:
1. 1 pull request per project. And each pull request should have 1 commit
2. I think we can ignore the lang files, as I understand the library wont break without them

abishekrsrikaanth:
Thanks, Let me update this and send you a fresh request

PR #3341, opened 2015-02-05
create colorbox 1.5.3

couponleaf:
added colorbox 1.5.3 version properly


jimaek:
This is an old version, why are you adding it?

couponleaf:
this version not available on jsdeliver so i want to add it

jimaek:
1.5.3 need to be the directory and inside the directory should be the actual file with the same filename as in the previous versions

PR #3379, opened 2015-02-08
Added refluxjs

csainty:


jimaek:
How about removing `reflux.js` from the PR and specifying the min file in update.json? I dont think we need the unminified file in production

csainty:
Done

PR #3413, opened 2015-02-12
Adding Library AmplifyJS 1.1.2

abishekrsrikaanth:
URL: http://amplifyjs.com/
Repo: https://github.com/appendto/amplify
Description: AmplifyJS is a set of components designed to solve common
web application problems with a simplistic API


files/amplifyjs/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "amplifyjs",
+  "repo": "appendto/amplify",
+  "files": {
+    "include": ["./lib/amplify.core.min.js","./lib/amplify.min.js","./lib/amplify.request.min.js","./lib/amplify.store.min.js"]
```
jimaek: Wrong. Use `basePath` to set `lib/` as root and then for include just use `*.min.js`

abishekrsrikaanth:
I will resend a new PR

PR #3414, opened 2015-02-12
Adding Auto Update for Bootstrap Switch Library

abishekrsrikaanth:
URL: http://www.bootstrap-switch.org/
Repo: https://github.com/nostalgiaz/bootstrap-switch

### DIFF
--- a/files/bootstrap.switch/update.json
+++ b/files/bootstrap.switch/update.json
@@ -1,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "bootstrap.switch",
+  "repo": "nostalgiaz/bootstrap-switch",
+  "files": {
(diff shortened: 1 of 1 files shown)


files/bootstrap.switch/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "bootstrap.switch",
+  "repo": "nostalgiaz/bootstrap-switch",
+  "files": {
+    "include": ["./dist/**/*"]
```
jimaek: This rule will create a `dist` directory which is wrong.
Instead use the `basePath` parameter to set `dist` and then you can remove `include` completely

abishekrsrikaanth:
Will resend a new PR. Closing this

PR #3415, opened 2015-02-12
Enabling Auto Update for jQuery.blockUI

abishekrsrikaanth:
Repo: https://github.com/malsup/blockui/


jimaek:
Cant be auto-updated because the repo doesnt have the min file that our hosted versions have

PR #3464, opened 2015-02-15
Adding Library csscritic 0.6.0

cburgmer:
Repo: https://github.com/cburgmer/csscritic
URL: http://cburgmer.github.io/csscritic


files/csscritic/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "npm",
+  "name": "csscritic",
+  "repo": "cburgmer/csscritic",
+  "files": {
+    "basePath": "dist/",
```
jimaek: I dont see a dist/ folder in the repo https://github.com/cburgmer/csscritic/tree/0.6.0

files/csscritic/update.json
```diff
@@ -0,0 +1,9 @@
...
+    "include": ["csscritic.allinone.js"]
```
jimaek: Same for the file, its not in the repo so the bot will fail to auto-update

cburgmer: Yes, but it is available from the npm package. Is this not enough? The build process generates the file and provides it in the package.

files/csscritic/update.json
cburgmer: See below

PR #3466, opened 2015-02-16
New library: dev-ribbon

brock:
Repo: http://github.com/brock/dev-ribbon
URL: http://brock.github.io/dev-ribbon


jimaek:
I dont think it makes sense to host this. This library was meant to work on localhost where access to the internet is not guaranteed. Also on localhost the performance is not important so the CDN is useless

brock:
No problem. I can understand your points. Thanks for providing an explanation.

PR #3473, opened 2015-02-16
update Cesium to 1.6

trehn:
I just took the contents of the 'Build' directory from http://cesiumjs.org/releases/Cesium-1.6.zip and put them into a new directory called '1.6'.

Will the 'latest' mechanism correctly consider 1.6 > b21?

@mramato any objections?


trehn:
As far as I can tell, these bot warnings are non-issues.

jimaek:
Thank you

PR #3498, opened 2015-02-18
Adding the latest version of GSAP and removing the oldest version

dfedgar:
Adding the latest version of GSAP and removing v12 because it's misleading, it's actually a really old version of GSAP.


dfedgar:
These changes are intentional.

jimaek:
We dont allow to remove or modify any existing files to ensure we dont break any websites.

dfedgar:
Ok, changed so that it's only adding the latest version of GSAP.

jimaek:
thank you

PR #3519, opened 2015-02-19
Add here-miss, dnt setting detection.

lewisgoddard:
I believe that just setting the mainfile means only that file is pulled, is that right?
Running libgrabber over this should result in version 1.1 being pulled in.


jimaek:
No, you need to add the update.json file to enable auto-updates and make the bot pull the files. Check out the readme

jimaek:
Can you please squash your commits into 1?

lewisgoddard:
I have tried and failed, unfortunately my internet is so slow it took over half an hour to get the repository.

jimaek:
You dont have to clone it. You can use the web interface to create 2 files resulting in 2 commits. But right now you have 3 commits.

PR #3520, opened 2015-02-19
Add slimmage 0.4.1

perosb:
adding slimmage


perosb:
Can auto update download from github releases?
If so is there an example of such a file?

In the slimmage case the .min.js is not packaged in anyway.

https://github.com/imazen/slimmage/releases/download/0.4.1/slimmage.min.js

jimaek:
Here you go https://github.com/jsdelivr/jsdelivr#auto-updating

perosb:
Thanks. Need to look at that later. The documentation only describes when the files are part of the dist tarball.
In this case it is not, only as separate download.

jimaek:
Do you deploy the min file to npm or bower? Our bot can work with that too

perosb:
I dont think so. I'm not the developer nor maintainer, but asked if it can be included in future releases.

Thanks!

jimaek:
Yeah unfortunately I dont see any other way. Its either github tags, npm or bower. Maybe add the min somewhere to we could enable auto-updates?

perosb:
Well, the min.js is added as a separate download in the tag/release: https://github.com/imazen/slimmage/releases/tag/0.4.1

Not sure if the bot can pull single files like that?
https://github.com/imazen/slimmage/releases/download/0.4.1/slimmage.min.js

jimaek:
This is what the bot sees https://github.com/imazen/slimmage/tree/v0.2.6

PR #3523, opened 2015-02-19
Adding lodash 3.2.0

jtmueller:
Adding lodash 3.2.0


jtmueller:
I do not understand the bot's objections at all. I'm just adding a missing version of an existing library. Why should "lodash.com" be expected to match "fronteers.nl" in the first place?

jimaek:
Looks like a bug, pinging @megawac to check it out.

megawac:
@jtmueller the `isMinimized` check isn't very clever, it literally just looks at line numbers. Maybe I should bold **Might be nothing** to make it clear that those checks are skeptical

jimaek:
@megawac What about the website thing? Im curious where fronteers.nl comes from

megawac:
Its something off about the certificate I suppose

megawac:
Looks like a temporary bug on the issuers end, the certificate was fixed

PR #3536, opened 2015-02-20
Added angular.ui-select plugin

sancospi:
Added angular.ui-select plugin


jimaek:
why 0.9.6 when the latest is 0.9.9?

sancospi:
Sorry, I'll update now

PR #3552, opened 2015-02-21
update update.json for trip.js

EragonJ:
as title.

### DIFF
--- a/files/jquery.trip.js/update.json
+++ b/files/jquery.trip.js/update.json
@@ -1,7 +1,7 @@
 {
   "packageManager": "github",
   "repo": "eragonj/trip.js",
-  "name": "Trip.js",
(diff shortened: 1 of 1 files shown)


EragonJ:
I did uploaded a new version on my repo, but I just noticed that the path is wrong because I did change it in my latest version. I am not quite sure this new version will be updated or not, hmmm !?

jimaek:
First of all I dont see it at all here https://github.com/EragonJ/Trip.js/releases

EragonJ:
@jimaek what you mean ? For me, all versions are under `https://github.com/EragonJ/Trip.js/releases`.

jimaek:
strange, when I checked there was no 3.0.0 on there

EragonJ:
@jimaek thanks !! :) ありがとう

PR #3571, opened 2015-02-23
Update project angular.bootstrap to 0.12.1

kevinoid:
This pull request adds the 0.12.1 release of UI Bootstrap from AngularUI. The files are the versions downloaded from https://angular-ui.github.io/bootstrap/ as the project does not keep built files in their repository.

As with the update to 0.12.0 in #2476, jsdeliverbot may complain about unescaped "«" and "»" characters, which I could replace if that would be preferable.

I hope you find it useful,
Kevin


jimaek:
Thank you

kevinoid:
My pleasure.  Thank you!

PR #3573, opened 2015-02-23
adding autocompeter

peterbe:
See https://autocompeter.com

PR #3574, opened 2015-02-23
autocompeter

peterbe:
see https://autocompeter.com

(this was originally opened https://github.com/jsdelivr/jsdelivr/pull/3573 but re-opened after a rebase)


jimaek:
Thank you

PR #3581, opened 2015-02-24
Update jquery ui to 1.11.3

perosb:


jimaek:
Given the multiple errors and the high profile of the project I think I will update it myself just to be sure nothing breaks.
Thank you for your help!

PR #3607, opened 2015-02-25
Adding the isotope-packery library

dfedgar:


jimaek:
Since this is an addon for isotope please use the format `isotope.packery` for the directory name

dfedgar:
Okay, the directory name has been changed to "isotope.packery".

PR #3677, opened 2015-03-01
update bxslider to v4.2.3, fix jsDelivr bot files, 3rd attempt

Aternus:


Aternus:
Dear JSDelivr, the plugin had been rewritten to comply with the latest standards thus the plugins folder have been renamed to vendor. Please allow this commit.

jimaek:
If you want to include everything inside dist/ then there is no need to use the `include` parameter. It will be added by default.
You also need to squash your commits into 1

Aternus:
I would rather be explicit about the files to include since there might be additional files and folders added later on.

What do you mean by squash my commits into 1?

jimaek:
Right now your PR has 8 commits, please squash them into 1. You can google how to do it.

Btw if you are going to update the update.json you can remove the actual files and let the bot upload them

PR #3697, opened 2015-03-02
ajile: auto-update now copies all files from /use/

iskitz:
* __updated: auto-update__ now copies all files from git repository's `/use/` directory.
* __updated:__ ajile developers can use their browser’s console to see what script(s) any version of ajile automatically loaded and understand how to modify that auto-load behaviour.


iskitz:
@jimaek: Please merge these changes. Here's my response to @jsdelivrbot's comment:

> There are some fundamental issues with your PR :no_good:
> * You appear to be changing the file contents of _index.js_ in _\<version-directory\>_!

The `index.js` script in each version directory is a helper script that __ajile__ automatically loads; it doesn't provide ajile APIs so modifying it doesn't negatively affect users. I've modified `index.js` to notify developers that it was auto-loaded by __ajile__ and to provide them with information about how to use or disable auto-loading.


> Looks like the info.ini has a few problems
> * Where is the ajile's mainfile (_com.iskitz.ajile.js_) in v-*0.9.1*

I've added that missing file. You can view it [here](https://github.com/iskitz/jsdelivr/blob/9099c27b462efa0060f8f094c65f77ec534a7cd2/files/ajile/0.9.1/com.iskitz.ajile.js) in this PR's changeset.


> We try to keep the file structure as consistent as possible between versions; there were some changes in file structure from previous versions...
> * In _ajile_ between _a.b.c_ and _x.y.z_ there was a change in file structure

There are no directory structure changes. I removed `com.iskitz.ajile.src.js` from the `0.5.5` directory because it was mistakenly added in the initial commit. The content of every other version directory is what it was in the original commit.


Thanks!

jimaek:
Please avoid updating existing files in the future. Since this change is not breaking and you are the author I merged it but in most cases we deny any updates.

iskitz:
Thanks! Will do.

PR #3716, opened 2015-03-03
MediaStreamRecorder

koko-ng:
I've add the library MediaStreamRecorder. It's a cross-browser implementation to record audio/video streams. I think it's useful because MediaRecorder is only supported by Gecko (cf. https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder)


jimaek:
I dont see a point of adding a `js` directory. Can you remove it and update the mainfile accordingly?
Also please add a `homepage` parameter and just link to Github.

Btw what source file did you use and how did you minify it?

pnommensen:
Ping @koko-ng.

PR #3777, opened 2015-03-08
Update angularjs to 1.3.14

as-com:


jimaek:
I believe the angular-scenario.js file is completely useless in production.

as-com:
Okay, removed it.

PR #3829, opened 2015-03-12
jquery lightslider

a5his:
Added jquery lightslider


a5his:
Thanks, will send another pull request.

PR #3830, opened 2015-03-12
jquery lightslider

a5his:
added jquery light slider


jimaek:
1. Rename the project to `jquery.lightslider`
2. I think its best to retain the original file structure. css/ js/ img/
3. Squash your commits into 1

PR #3832, opened 2015-03-12
My plugin

himanshusharma1:


himanshusharma1:
How can I merge it??

files/kalturahtml5plugin/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Himanshu Sharma"
+github = "https://github.com/himanshusharma1"
```
jimaek: This needs to be a link to the project itself

pnommensen:
The use case here seems very limited and I do not see the source files. Can you link them and explain where it's being used? Thanks.

pnommensen:
ping @himanshusharma1

PR #3913, opened 2015-03-18
Add auto-updating to sly

lewisgoddard:
I noticed sly was quite out of date, and seeing as they ship quite often this seems like a good idea. I didn't bother including the uncompressed version.


jimaek:
Thank you

PR #3922, opened 2015-03-19
Add missing Highcharts v4.1.4

spektom:


spektom:
The check is wrong! Even though the content of these files was not empty, it was erroneous: sometimes it was empty, sometimes it was from the previous version.

jimaek:
Strange, thank you for fixing this.
I will try to make sure it wont happen again

PR #3965, opened 2015-03-23
Add github cards.

lepture:
Approve required.

https://github.com/lepture/github-cards/issues/26


lepture:
@jsdelivrbot Actually, we need those html files.

PR #3979, opened 2015-03-24
added php melody 

topvd:
i have added php melody template files


topvd:
the files are for my template i don't know how chose the mainfile

PR #3982, opened 2015-03-24
add XEIcon

bnu:


pnommensen:
Please squash your commits, and also the non minified file is not necessary for production use. Thanks.

PR #4014, opened 2015-03-26
Add project jquery.planit at 1.0.1

seancdavis:
Planit is a jQuery plugin for managing markers/hotspots on top of an image or within a blank canvas. This is the latest version.


jimaek:
Thank you

PR #4034, opened 2015-03-27
Changing to the new GitHub development  site

### LINKED ISSUE #4033: Change Github reference for Ramda
I'm not sure if this is the correct way to go about this, but I could find nothing in the actual repo that would let me set this.

On the Ramda page of JSDelivr (http://www.jsdelivr.com/#!ramda) the GitHub link points to the original development site for the project, when it was part of my personal account.  It's moved to a group account now, and I was hoping that it would be easy to change that link to https://github.com/ramda/ramda.  Is that something I can control myself, or is an issue like this the right way to do so?

Thanks.

CrossEye:
Closes #4033

### DIFF
--- a/files/ramda/info.ini
+++ b/files/ramda/info.ini
@@ -1,5 +1,5 @@
 author = "Ramda Folks"
 homepage = "http://ramdajs.com"
-github = "https://github.com/CrossEye/ramda"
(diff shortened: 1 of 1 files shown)


jimaek:
Thank you

CrossEye:
Thanks for your help

PR #4046, opened 2015-03-28
Add Hover.css

Aquei:
A collection of CSS3 powered hover effects to be applied to links, buttons, logos, SVG, featured images and so on.  http://ianlunn.github.io/Hover/


jimaek:
Thank you

PR #4057, opened 2015-03-29
time sequence steps, drop .map encouragement

tomByrer:
Great idea to have a CONTRIBUTING doc!
Since you have to get files ready first before switching on the bot or uploading, I added those instructions first.
As per discussion with @as-com &/or @megawac (sorry forgot, few months ago), we decided that `.min.map` files don't belong on a public CDN.  They are used help live-edit code, which is not recommended to do on a production site.  They can host non-minified code if they need to edit their live site.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,13 +1,23 @@
+### Preparation
+
+1. Choose your files for public hosting:
+ * Do not upload useless files like demos, examples, licenses, readmes and any other files not being used in production.
+ * Please add **only the minified** versions for all files if you are adding a project for the first time.  Also compress images & fonts the best you can.
+ * If there is a previous version of the project you are adding, please ensure that the new version contains same files. For example if in the previous version there are `*.min.js` and `*.js` (& perhaps `*.min.map`) files, please add both to the new version.
+2. Ensure there is a open-source or 'free to use commercially' license.
+3. (Optional: recommended) Ensure you have a [comment banner](https://github.com/tomByrer/web-engeneering-notes/blob/master/banner-tips.md#mini-banner) in your files' top line.
+4. Choose adding your project via GitHub's [web interface](#web-interface-recommended) or [`git`](#advanced).
+
+
 ### Web interface (Recommended)
 
 *This is limited to projects that can be auto-updated by our bot. It requires github/npm/bower with tagged versions.*
 
-1. Using your browser go to [files/][1] and use [Github's interface](https://cloud.githubusercontent.com/assets/1834071/6826939/4019ce7a-d30d-11e4-8d1b-7821b923dd50.gif) to create a new folder with an info.ini file inside. [Documentation](https://github.com/jsdelivr/jsdelivr/blob/master/CONTRIBUTING.md#file-structure)/[Example][2]
-2. In the same directory create an update.json file to enable auto-updates. [Documentation](https://github.com/jsdelivr/libgrabber#add-updatejson-schema)/[Example](https://github.com/jsdelivr/jsdelivr/blob/master/files/angular.moment/update.json)
+1. Using your browser go to [files/][1] and use [Github's interface](https://cloud.githubusercontent.com/assets/1834071/6826939/4019ce7a-d30d-11e4-8d1b-7821b923dd50.gif) to create a new folder with an info.ini file inside. [Documentation](https://github.com/jsdelivr/jsdelivr/blob/master/CONTRIBUTING.md#file-structure) / [Example][2]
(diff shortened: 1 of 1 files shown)


jimaek:
most min.js files include a .map file in their source code that results in the browser sending that request anyway. So unless we have a way to remove the .map URL from .min.js then I believe we need to keep uploading the map files.

megawac:
Yeah completely disagree. .min.map is a good fallback in case something
goes wrong

tomByrer:
Most?
A few popular JS do have .map files, but most of the ones I've been finding don't.  Half don't even include all the files minified.

The .map reference *may* be built, but IIRC must minifers don't automatically build the map.
Worst that happens without a map that is expected is 404s when browser dev tools are opened & a get complaints on SO.  No one noticed the were missing from jQuerry when I made that PR; I did it because I had a moment of perfection.  Not so worried any more.

tomByrer:
> Yeah completely disagree. .min.map is a good fallback in case something goes wrong

I don't want to stir up a debate, but I am truly interested in finding out how many big (& not so big, like 1000 hits/day) websites would hot-edit their JS on a production server?  I would think most would edit offline & push after testing.

I could see live editing something on Codepen or jsFiddle, but you could serve uncompressed files via https://rawgit.com/  They can mirror GitHub files on this great CDN called [MaxCDN](https://www.maxcdn.com/), you might have heard of it?  ;)

jimaek:
Although I like your changes I don't like how you made "Advanced" look easier than the web method.
Maybe we can add more instructions there? Because it looks too small right now.

Regarding maps I agree that they are useless but a fair amount of projects still uses them. I dont think its our place to define coding standards. If the source code loads the map file then I would prefer to have it hosted. It doesnt take much space, it ensures there are no errors and it is what the author wanted.
Plus we already have so many of them that its too late to stop adding them.

CONTRIBUTING.md
```diff
@@ -1,15 +1,25 @@
+### Preparation
+
+1. Choose your files for public hosting:
+ * Do not upload useless files like demos, examples, licenses, readmes and any other files not being used in production.
+ * Please add **only the minified** versions for all files if you are adding a project for the first time.  Also compress images & fonts the best you can.
+ * If there is a previous version of the project you are adding, please ensure that the new version contains same files. For example if in the previous version there are `*.min.js` and `*.js` (& perhaps `*.min.map`) files, please add both to the new version.
+2. Ensure there is a open-source or 'free to use commercially' license.
+3. (Optional: recommended) Ensure you have a [comment banner](https://github.com/tomByrer/web-engeneering-notes/blob/master/banner-tips.md#mini-banner) in your files' top line.
```
megawac:
```diff
- (Optional: recommended) Ensure you h
+ (Optional: recommended) H
```

CONTRIBUTING.md
```diff
@@ -44,13 +53,8 @@ A project's directory should contain the following:
 A version directory should contain the following:
 
 1. Static files needed for the project to work.
-2. If there is no minified version of the main JS/CSS file, please create your own using this ([minification tool][3]).
-3. If there are official or expected source maps for the minified js, please include those in the folder.  Currently, the following projects officially support the `.map` files:
-  * angularjs
-  * jQuery
-  * mithril
-4. Do not upload useless files like demos, examples, licenses, readmes and any other files not being used in the production.
+2. To shrink JS/CSS, you can use YUI ([minification tool][3]).
```
megawac: Can we change this to uglify2

tomByrer: Uglify only does JS, YUI does both (though Uglify2 is better for JS).
Is there a web interface that will compress all?

tomByrer:
> you made "Advanced" look easier than the web method

I don't know what you mean; I added only 4 spaces in the web method.  Perhaps you are thrown off by my new "Preparation" section (which is mostly your information moved before other steps)?  Preparation is to be done before either web or git methods.

CONTRIBUTING.md
as-com: http://refresh-sf.com

megawac: Nice

tomByrer: Seems very good @as-com!  Thought does drop top-line-comment, which is important for asset tracking & copyrights, very important since we contaminate files.
Maybe I should fork?

megawac: Or just send em a PR with that option enabled by default

jimaek:
Exactly, web has detailed instructions in 5 steps while git has very little text in just 2 steps. It makes it look the easiest and fastest way. We need to either remove the prepare part or heavily improve the git part

CONTRIBUTING.md
as-com: That website was the website @jimaek was linking to all the time. :smile:

tomByrer:
"Preparation" is meant to be used for *both* web & git, since they both overlap there.
Should we number the headers to make that more clear?

PR #4097, opened 2015-04-01
add mojs library

legomushroom:


jimaek:
Perfect, thank you

legomushroom:
Thank you! When can I expect to see it in the search?

jimaek:
The files get uploaded instantly but the search might take a little longer.
Its there now http://www.jsdelivr.com/#!mojs

PR #4113, opened 2015-04-02
Add SPDY note

megawac:


jimaek:
SPDY and HTTP2 are different things. Our CDNs only support SPDY based on Nginx's spdy module

megawac:
Ah drat - I'll amend that

pnommensen:
the performance with SPDY for single/few static file delivery is questionable.

jimaek:
Yes, but we want to make sure people know we support SPDY so that if they use SPDY on their website jsDelivr wont be the slow player.

PR #4215, opened 2015-04-07
removed ion.rangeSlider, fixed old one

zippy1981:
The rules would seem to indicate the images should have been pulled. However, they weren't. So, I don't know whats happening.


megawac:
calling @jimaek

jimaek:
yeah, its good. We just need a way to hide projects from the API (both v1 and v2)

PR #4227, opened 2015-04-08
Add socialmedia.js

jabranr:
Thank you in advance! :+1:

### DIFF
--- a/files/socialmedia/update.json
+++ b/files/socialmedia/update.json
@@ -1,0 +1,8 @@
+{
+	"packageManager": "github",
+	"name": "socialmedia",
+	"repo": "jabranr/socialmedia",
+	"files": {
+		"include": ["socialmedia.js", "socialmedia.min.js"]
+	}
+}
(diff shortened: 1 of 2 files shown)


pnommensen:
Hi @jabranr the non minified file doesn't seem necessary for production? If this is the case please remove it. thanks.

jabranr:
Hi @pnommensen, Sure I can. What if I add source map file along, would that make any sense to keep the non-minified file for debugging? Thanks.

files/socialmedia/update.json
```diff
@@ -0,0 +1,8 @@
+{
+	"packageManager": "github",
+	"name": "socialmedia",
+	"repo": "jabranr/socialmedia",
+	"files": {
+		"include": ["socialmedia.js", "socialmedia.min.js"]
```
jimaek: you need to use `basePath` to set the directory to `dist/` because the included files are there https://github.com/jabranr/socialmedia/tree/master/dist

jabranr: Hi @jimaek, I have added a `basePath` to `/dist` and removed non-minified JS as requested above. Cheers

jimaek:
Thank you

PR #4235, opened 2015-04-08
First using jsdelivr

temandoedoek:
This is my first try. just want to know whether whats i doing right following instructions.tank.


files/coba/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "temandoedoek"
+github = "https://github.com/temandoedoek"
```
jimaek: This needs to be the link to the project itself

files/lagi/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "temandoedoek"
+github = "https://github.com/temandoedoek/lagi"
+homepage = "http://darno.tk/"
+description = "Still confuse using jsdelvr"
+mainfile = "lagi.min.js"
```
megawac: Is this file not included?

megawac:
Good start here, just need to a) do one pull request per library b) use a meaningful home page with documentation/or github.

Also the two libraries you submitted don't really have documentation on the github pages which is where you should start (or submit for different libraries)

Thanks

PR #4262, opened 2015-04-10
Added jquery.lazy

eisbehr-:


pnommensen:
Hi @eisbehr- the non minified file isn't necessary for production, right? We generally only include the minified files. Please remove that file and then squash the commits. Thanks :)

jimaek:
Looks good but we allow a max of 2 commits per PR. Please either squash your commits or create a new PR

eisbehr-:
created a new pr

PR #4284, opened 2015-04-13
Add update.json for isotope

hubdotcom:
Automatically update to latest releases of isotope

### DIFF
--- a/files/isotope/update.json
+++ b/files/isotope/update.json
@@ -1,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "isotope",
+  "repo": "metafizzy/isotope",
+  "files": {
+    "basePath": "dist/"
+  }
+}
--- a/files/isotope/update.json
+++ b/files/isotope/update.json
@@ -3,6 +3,6 @@
   "name": "isotope",
   "repo": "metafizzy/isotope",
   "files": {
-    "basePath": "dist/"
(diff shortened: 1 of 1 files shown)


pnommensen:
Looks like 'jquery.isotope.js' was included in all previous versions. Users who are using that file in production will face errors if they are using the /latest/ feature. 

@jimaek  thoughts?

hubdotcom:
jquery.isotope.js is still included, it's in the /js directory, will update the json to reflect this

hubdotcom:
@pnommensen how's this? https://github.com/hubdotcom/jsdelivr/commit/a4b68cb42dc5db6da82f31ad1cf885aabd10049c

jimaek:
isotope.js is not the same as jquery.isotope.js

If you dropped the jquery from the name completely then there is nothing we can do, you can remove the js file and leave only dist/

hubdotcom:
okay, so we are good to go with this?
```
{
  "packageManager": "github",
  "name": "isotope",
  "repo": "metafizzy/isotope",
  "files": {
    "basePath": "dist/"
  }
}
```

jimaek:
Yes, looks good

hubdotcom:
Single commit PR at https://github.com/jsdelivr/jsdelivr/pull/4292 - is this the preferred method of commiting to jsdelivr?

jimaek:
Yes, we have a max of 2 commits per PR

PR #4287, opened 2015-04-13
Added timezonedb.min.js

seikan:


pnommensen:
Thanks @seikan. Please squash your commits. It's also not necessary to include files/timezonedb/1.0.0/timezonedb.min.js -- a good way to make sure that the update.json is working as expected.

seikan:
Thanks. I will fix this.

PR #4295, opened 2015-04-13
Add update.json for jsdelivr

hubdotcom:


jimaek:
Thank you

hubdotcom:
My pleasure, it would be great if we could see some stats - the top 10 projects without an update.json would be ideal for contributors to focus on the most active projects

PR #4383, opened 2015-04-18
Added Jdenticon

dmester:
Added info.ini and update.json for Jdenticon


dmester:
Will recreate this request

PR #4432, opened 2015-04-21
Add update.json for Responsive Nav

hubdotcom:


megawac:
Thanks for these @hubdotcom!

hubdotcom:
No problem, good to get the update.json's in place - need to have a reputation for the fastest AND most up to date CDN!

jimaek:
@hubdotcom navigation-nav doesnt include the min file in the latest version https://github.com/viljamis/responsive-nav.js/tree/v1.12

So the update.json wont work, the bot already failed https://github.com/jsdelivr/jsdelivr/pull/4434

viljamis:
Noticed this popping up in my notifications. v1.12 is actually two years old already, current latest version is 1.0.39. Seems I’ve switched the way the releases are numbered and tagged at some point after releasing the plugin. So, what I did now, is that I went through the whole history of releases and cleaned up everything manually so there’s no more v1.12 at all and 1.0.39 should now be the latest.

Sorry about this!

jimaek:
@viljamis Awesome, thank you!

hubdotcom:
Fantastic!

PR #4439, opened 2015-04-22
Update zepto to 1.1.6

as-com:
I made a build with all the modules, rather than having them all in separate files. (The ones in the modules folder are just empty files)

I felt that it would be easier to maintain with the modules all built into one file.


as-com:
Huh. The bot thinks that empty files are binary.

jimaek:
why are they empty?

as-com:
The JS is already in the main file. The empty files are placeholders.

jimaek:
What do you mean? Whats the point of them if they are empty?

as-com:
To avoid 404’s when people use /latest/

jimaek:
But their code wont work anyway no matter 404 or 200 since he files are empty

as-com:
What I did here is put all of the code that would have been in those individual modules into one combined file. Zepto does not provide individually minified versions of the modules, so I used their build tool to create a minified version of Zepto that contains all of those modules.

jimaek:
I just dont see how the empty files help. If someone links to it it wont work. Am I missing something?

pnommensen:
I don't quite understand this either.

PR #4442, opened 2015-04-22
Adding missing compat build.

jtmueller:
Is there any way to block approval of an update that is missing 50% of the files that were present in the previous version? I can't very well use the jsdelivr "most recent minor version" feature if the most recent minor version is likely to just drop the IE8-compatible build entirely.


jimaek:
Well its up to each developer. We cant decline an update because of some decisions the author himself made. When an update that is missing files arrives it is manually reviewed. 
If required we create the correct mainfile overrides and updates to info.ini to ensure /g/ functionality doesnt break.
But we cant do anything about anything else. 
If you have any ideas let me know.

Regarding this PR. lodash removed compat builds from his repo https://github.com/lodash/lodash/tree/3.7.0 so I dont think it makes sense to manually add these files every single time.
lodash is auto-updated by our bot

megawac:
@jimaek @jtmueller the way cdnjs is handling auto-updating for lodash is to have a `lodash` and `lodash-compat` repository. Does that make sense for us to do as well (will break `http://cdn.jsdelivr.net/lodash/3/lodash-compat.js` links

@jtmueller see https://github.com/lodash/lodash/issues/1107

jtmueller:
The official lodash site still has a compat build for 3.7.0 so I don't think it's correct to say that compat builds have been removed.

However, I'm find if jsdelivr wants to consider lodash-compat a separate library. It means I won't have to specify a filename in parens on the jsdelivr URL anymore.

I'm not concerned about it breaking http://cdn.jsdelivr.net/lodash/3/lodash-compat.js type links, because those links have been broken ever since someone added lodash 3.6 without any compat files 11 days ago. If putting the compat build in its own folder on jsdelivr means it won't suddenly vanish on me, and will have timely updates, I'm all for it!

jimaek:
We always try to avoid removing files between minor versions but some times its impossible to avoid.
We have more than 1500 projects here and they all change format, names, code...

As much as we try its impossible when the author hardly decides on something. 

Regarding this case I am cool with adding lodash-compat as `lodash.compat` to jsDelivr and enabling auto-updates for it. Who wants to do it?

jtmueller:
@jimaek I can submit a pull request with the files, but I have nothing to do with either lodash or jsdelivr, and I have no idea how to set up auto-updates. I am not prepared to become the official go-between with a project I'm not involved in and a CDN I've barely used.

megawac:
All that is needed is a copy of the update.json and info.INI files to the
lodash xompat folder

jimaek:
You dont need to become a go-between for anything. Its a one time thing to add the project and enable the updates, the rest will be taken care by the bot and us.
If you dont figure it out I can do it myself a bit later

jtmueller:
I'll give it a try and submit a new pull request for that.

PR #4474, opened 2015-04-24
Add Esri Leaflet

patrickarlt:


jimaek:
Since this is an addon for leaflet please rename the directory to `leaflet.esri`.
Also is `esri-leaflet.js` a minified file? If not can you set the min version there?

patrickarlt:
@jimaek Going to close and reopen with the proper directory name this since its taking forever to pull down the repo so i can squash commits.

Yes `esri-leaflet.js` is a minified file. I'm just following leaflets pattern for minified/unminified file names.

PR #4296, opened 2015-04-13
Update project imagesloaded to v3.1.8

jsdelivrbot:
libgrabber updated project imagesloaded to v3.1.8.

@desandro

### Metadata information

```json
{
  "metadata": {
    "packageManager": "github",
    "name": "imagesloaded",
    "repo": "desandro/imagesloaded",
    "files": {
      "include": ["imagesloaded.js", "imagesloaded.min.js", "imagesloaded.pkgd.js", "imagesloaded.pkgd.min.js"],
      "exclude": []
    },
    "path": "/home/jsdelivr/files/imagesloaded",
    "localVersions": ["2.1.0", "3.0.1", "3.0.4", "3.1.0", "3.1.4", "3.1.6"],
    "remoteVersions": ["v1.0.0", "v1.0.1", "v1.0.2", "v1.0.3", "v1.0.4", "v1.1.0", "v1.2.0", "v1.2.1", "v1.2.2", "v1.2.3", "v1.2.4", "v2.0.0", "v2.0.1", "v2.1.0", "v2.1.1", "v2.1.2", "v3.0.0", "v3.0.1", "v3.0.2", "v3.0.3", "v3.0.4", "v3.1.0", "v3.1.1", "v3.1.2", "v3.1.3", "v3.1.4", "v3.1.5", "v3.1.6", "v3.1.7", "v3.1.8"],
    "branchVersions": []
  },
  "version": "v3.1.8",
  "updated": true,
  "updatePath": "/home/jsdelivr/files/imagesloaded/3.1.8",
  "branch": "imagesloaded/v3.1.8"
}
```

PR #4491, opened 2015-04-25
Update imagesloaded to latest

hubdotcom:
Apologies for the 4 commits - didn't want to clone the entire repo to get this added


jimaek:
What is the source for these files? The repo doesnt have the min file anymore

hubdotcom:
Compressed with UglifyJS2 default settings via http://refresh-sf.com/

Author has stated he no longer sees any point to support the imagesloaded.min.js - he said it's useless without the packages so only the pkgd min will be uploaded (#4296).

jimaek:
@hubdotcom Just add the min file. I merged the other files from the bot's PR

hubdotcom:
Should we still add the imagesloaded.min.js or should we phase this out?

pnommensen:
There should always be a min file. Mind asking the author of this lib to include in the packaged build?

hubdotcom:
I did - see thread above. Author says the non packaged file is useless without the dependencies so a min file would not be suitable for a production environment

hubdotcom:
Author changed the name of the core file from the .min.js to the .pkdg.min.js in v3.1.5. Anyone that continues to use the /latest/ min.js will have their scripts break anyway

PR #4492, opened 2015-04-25
Update to current year

hubdotcom:
Same convention followed by:
https://github.com/twbs/bootstrap/blob/master/LICENSE
https://github.com/angular/angular.js/blob/master/LICENSE


jimaek:
Thanks

PR #4508, opened 2015-04-27
Lodash compatibility build

jtmueller:
I'm not sure if I set up info.ini and update.json correctly for automatic update or not.


files/lodash.compat/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "bower",
+  "name": "lodash.compat",
+  "repo": "lodash/lodash",
+  "files": {
+    "include": ["**/*.js", "**/*.map"]
```
jimaek: I dont see any map files in there. I think a simple "include": ["lodash.min.js"] would be enough, you can even remove the unminified file from this PR.
What do you think?

jtmueller: I was just copying the file from the non-compat lodash folder. Since that version included both minified and non-minified, I made this one the same. I don't have a problem with the changes you suggested, did you want me to apply them?

jimaek: yes, you can simply change it to `["lodash.min.js"]`

jtmueller:
Yes? No?

PR #4540, opened 2015-04-29
Add jsblocks

astoilkov:


files/blocks/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "blocks",
+  "repo": "astoilkov/jsblocks",
+  "files": {
+    "basePath": "dist/"
+  }
```
jimaek: add also an include parameter for:
"*.min.*", "./mvc/*.min.*", "./query/*.min.*"

astoilkov: Is this going to include all files in all directories. Including the .js and .map files in `mvc` and `query` folders?

jimaek: It will include only the min files from dist, mvc and query

astoilkov: Can we do:
```javascript
{
  "packageManager": "github",
  "name": "blocks",
  "repo": "astoilkov/jsblocks",
  "files": {
    "include": ["*.min.*", "./mvc/*.min.*", "./query/*.min.*", "*.map", "./mvc/*.map", "./query/*.map", "./mvc/*.js", "./query/*.js"],
    "basePath": "dist/"
  }
}
```

jimaek: Why do you want the unminified files?

astoilkov: So people can take a look at the source code. Howerver, it is not that important. I could leave them out if you suggest to do so.

jimaek: Well, jsDelivr is for production websites. Source code is not relevant there, the most important is speed. My original config should work fine for that if you agree

astoilkov:
I agree with you. You could take a look at the final version.

PR #4547, opened 2015-04-29
Updated to 1.3.15

sancospi:
Updated to 1.3.15


sancospi:
- It's throwing a problem in a version I didn't modify.
- angular.min.js is the official minimized version downloaded from angular's site
- angular-scenario.js is a new file, not a deletion

jimaek:
angular-scenario.js is not needed

PR #4549, opened 2015-04-29
jdlib JavaScript Library

jackdalton:
A library of useful JavaScript functions.


jimaek:
You dont have any tagged versions. Also the lib looks pretty custom to be used by only you. Most people would only need to copy a few functions from it instead of loading the whole files just for 1 function.
I dont think if it makes sense to host it.

jackdalton:
@jimaek Fair enough. I'll look elsewhere.

PR #4551, opened 2015-04-29
Fixed typed.js

fatbattk:
v1.1 format.


jimaek:
Why no more min files?

fatbattk:
There never was a min file below 1.1 release. Min file was added to `master` later but no new release yet.

jimaek:
Can we get the author to include it? Because for production we need the min file

megawac:
It looks like its fine as is; feel free to comment otherwise
https://github.com/mattboldt/typed.js/tree/master/dist

PR #4613, opened 2015-05-04
Create info.ini

tom2strobl:

PR #4614, opened 2015-05-04
Create update.json

tom2strobl:


pnommensen:
@tom2strobl: The non-minified file doesn't seem necessary for production use. If this is the case, please update the update.json accordingly. Thx.

tom2strobl:
Updated both files to be in the same PR and excluded the non-minified version as you're right, nobody will really need it.

tom2strobl:
PR #4613 is probably now to be closed, right?

files/audiofx/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Thomas Strobl"
+github = "https://github.com/madebywild/audioFX"
+homepage = "https://github.com/madebywild/audioFX"
+description = "High-Level Audio Effects using the Web Audio API in JavaScript, ~1-2 KB gzipped."
+mainfile = "dist/AudioFX.min.js"
```
jimaek: Just `AudioFX.min.js` is the correct

jimaek:
Once you fix the problems please either squash your commits or open a new PR. The maximum amount of commits per PR is 2

tom2strobl:
Closed and opened #4628

PR #4654, opened 2015-05-06
Added HTML Kickstart

ethanpil:


files/HTML-KickStart/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Joshua Gatcke"
```
megawac: Change folder to `html-kickstart`

jimaek:
We already host it over here http://www.jsdelivr.com/#!99lime

ethanpil:
99Lime is the name of the developer not the name of the library, which is why i couldnt find it when i searched...  the library is called HTML Kickstart. Perhaps a rename would be in order?

PR #4773, opened 2015-05-14
Add sockjs update.json

brycekahle:


files/sockjs/update.json
```diff
@@ -0,0 +1,9 @@
+{
+	"packageManager": "github",
+	"name": "sockjs-client",
+	"repo": "sockjs/sockjs-client",
+	"files": {
+		"basePath": "dist/",
+		"include": ["sockjs.min.js", "sockjs.js"]
```
jimaek: I think we can only include the min file. Can you remove the unminified one please?

brycekahle:
Updated to only include the minified file.

jimaek:
Thank you

PR #4809, opened 2015-05-18
ng-file-upload

akkumar:
Lightweight Angular JS directive to upload files - 4.2.1 version added.


jimaek:
We already host this project http://www.jsdelivr.com/#!angular.file-upload
You can update it if you want or add an update.json file

akkumar:
Seems like the author had renamed the package since then . ( The info.ini link for angular-file-upload redirects to the new name ) 

And the filenames are renamed as well.  

So - submitted another pull request here  -  https://github.com/jsdelivr/jsdelivr/pull/4822 . 

jsdelivr bot is not happy for the reason mentioned above.  Let me suggest what are the options now  ? ( a new package vs existing package ) ?

PR #4822, opened 2015-05-19
Ngfileupload

akkumar:
Adding version 4.2.1 per earlier suggestion. 

The file names of the distribution seems to have changed though. 

Also since then, the author has renamed the project in github as well. 

Please suggest what the right thing to do ( new package vs old package )


jimaek:
@jsdelivrbot squash and merge

PR #4842, opened 2015-05-19
Fastclick version update 1.0.4 to 1.0.6

madssj:
Added FastClick versions from 1.0.4 to 1.0.6 with their own `.min.js` file, which is ~2.5 KiB smaller, than the previously provided.


madssj:
Uhm, well, I didn't touch that version, but fair enough, I'll take one for the team.

jimaek:
Why all the versions and not just the latest one?

madssj:
@jimaek for completeness. Someone might really need `1.0.5` and it'd be a shame to disappoint them.

jimaek:
@jsdelivrbot squash and merge

jimaek:
Thank you

PR #4854, opened 2015-05-20
Updated js and css of wp-jupiter

elihuvillaraus:
An updated version from 2.x to 4.x of wp-jupiter css and js


elihuvillaraus:
I do not understand why automatic validation is showing errors on everyfile saying that there´s no new line at the end of file

jimaek:
The whole PR is incorrect. Its also not the best idea to host this kind of files

PR #4883, opened 2015-05-21
Add AdminLTE

irazasyed:
Project: https://github.com/almasaeed2010/AdminLTE

### DIFF
--- a/files/admin-lte/info.ini
+++ b/files/admin-lte/info.ini
@@ -1,0 +1,5 @@
+author = "Abdullah Almsaeed"
+github = "https://github.com/almasaeed2010/AdminLTE"
+homepage = "https://almsaeedstudio.com/"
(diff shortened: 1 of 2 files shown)


jimaek:
Sorry but we cant host the files of this theme. You are welcome to upload individual projects that the theme uses, for example we already have bootstrap and jquery.

irazasyed:
@jimaek Is there a specific reason for not hosting? It's an open-source and highly popular project.

Also, If you notice the `update.json` file, I've only included project specific files and not other dependencies (No bootstrap, jquery or any other external libraries), they're not at all included. So it's like hardly 5-10 files (Including skins).

jimaek:
Its open source but its mean to be customized, meaning most people will edit the source code to make it look like they need

PR #4938, opened 2015-05-23
Added cookieBAR

ToX82:


jimaek:
You are completely missing the `info.ini` file. Please check the README

ToX82:
My bad, it should be there now

jimaek:
If you are going to add almost every single file in that directly then instead of include use exclude to exclude LICENSE, README and the others.
It will also fix an error in your include parameter

jimaek:
@jsdelivrbot squash and merge

PR #4951, opened 2015-05-24
OWNER: Update textAngular File structure.

SimeonC:
Fixed my confusing file structure so everything is in the correct folders, now all files we need are in the dist folder, this includes non-minified files though, is that OK or would you rather I exclude them?

### DIFF
--- a/files/angular.textangular/update.json
+++ b/files/angular.textangular/update.json
@@ -4,7 +4,6 @@
   "repo": "fraywing/textAngular",
   "files": {
     "include": [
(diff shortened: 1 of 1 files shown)


jimaek:
I see you also added this file `textAngularSetup.js` is it required in production?

SimeonC:
Only if you're running the unminified version, which some people will do. (Whether they should is a different matter).

jimaek:
OK, thank you

PR #4963, opened 2015-05-26
Add AnchorJS to jsdelivr.

bryanbraun:
Closes bryanbraun/anchorjs#25.


pnommensen:
@bryanbraun Could you remove the non-min file? Not necessary for production (CDN) use.

bryanbraun:
I've removed the non-minified file.

PR #4971, opened 2015-05-26
info.ini

girishakj:
This is to host angular-rickshaw libraries on jsdelivr cdn.


jimaek:
1 PR per project please https://github.com/jsdelivr/jsdelivr/pull/4970

PR #4973, opened 2015-05-26
angular-rickshaw library

girishakj:
Request to host angular-rickshaw library on jsdelivr


pnommensen:
@girishakj: If you don't create an update.json we can't update this automatically. Please see readme for details.

girishakj:
I have added update.json, would this work now?

girishakj:
Looks like, its not working, will resubmit.

PR #4975, opened 2015-05-26
Fix handlebars.js mainfile

kokarn:


jimaek:
If we apply this fix then we need to fix the older versions that dont have a `handlebars.min.js` file.
We need to create inside each version folder a file called `mainfile` that contains the filename `handlebars.runtime.min.js`

kokarn:
Don't you mean that the ```mainfile``` file should contain ```handlebars.min.js``` or is this some other cool voodoo?

I've done another "fix" for this, copying the ```handlebars.js``` file in each folder to ```handlebars.min.js``` since i was unable to minify ```handlebars.js```. 

Let me know how you want this fixed and I'll happily do it.

jimaek:
Here an example of what I meant https://github.com/jsdelivr/jsdelivr/blob/master/files/imagesloaded/2.1.0/mainfile 
But your way works too in this case

kokarn:
Ah, nice example.

Thank you so much for updating this! Has been driving me nuts more than once :)

PR #5017, opened 2015-05-28
Create update.json

tivie:


tivie:
Thanks @jimaek for the quick merge.

Regarding info.ini, the file is outdated. Shall I create a new PR with the changes?

And the `mainfile` directive... should it point to `dist/showdown.min.js` or just `showdown.min.js`?
(obviously the distribution files live in the dist folder)

jimaek:
You can create a new PR if you want. The mainfile should be showdown.min.js, dist/ gets dropped by the bot

PR #5024, opened 2015-05-29
Create info.ini

RoberMac:


jimaek:
Both files need to be in the same PR. Dont do a new pr for each file
https://github.com/jsdelivr/jsdelivr/pull/5025

PR #5049, opened 2015-05-30
Update project babylon.js to 2.1-beta

vousk:
We have finally choosen the right stable 2.1 beta file to put on CDN:
Source file is https://github.com/BabylonJS/Babylon.js/blob/4a3be3405d5679449dde1ccce2f1849f2dec2e07/babylon.2.1-beta.js


jimaek:
Where is that file in the actual repo?

vousk:
This file is in commit 4a3be3405d5679449dde1ccce2f1849f2dec2e07 of the actual repo.
We are updating the beta file at (almost) each commit, but the latest beta file cannot be put in CDN because it is not enough tested in production environnement, that's why the candidate for CDN always comes from an older commit.

PR #5061, opened 2015-06-01
AlaSQL added

mathiasrw:
Added AlaSQL ( www.alasql.org )

### DIFF
--- a/files/alasql/update.json
+++ b/files/alasql/update.json
@@ -1,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "alasql",
+  "repo": "agershun/alasql",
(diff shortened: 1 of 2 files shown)


files/alasql/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "alasql",
+  "repo": "agershun/alasql",
+  "files": {
+    "include": ["./dist/*"]
```
jimaek: `"include": ["dist/*.min.js"]` should be enough

mathiasrw: :+1:

PR #5103, opened 2015-06-03
webcomponents 1.0.0 

akkumar:


akkumar:
Ok - So I am only adding a new version of lodash here , as needed by bower_components of webcomponents. 

I am not changing the info.ini of lodash as part of this commit. 

The message reported is - "hostname 'lodash.com' doesn't match either of '*.fronteers.nl', 'fronteers.nl'" . 

Can someone suggest what needs to be done here ? Thanks !

akkumar:
For now, I have removed lodash from this update altogether.

jimaek:
You say you added v1.0.0 but the files inside report 0.7.2. Are you sure you are using the correct files?

jimaek:
I dont think v1.0.0 even exists https://github.com/webcomponents/webcomponentsjs/releases

akkumar:
You are right. Sorry to have confused this with polymer release 1.0.0 . Closing this for now.

PR #5105, opened 2015-06-03
loadCSS

akkumar:


jimaek:
Only the directory needs to be lowercase, the file itself is best to have the original name

akkumar:
File renamed as per original name (leaving lowercase for dir only).  Let me know if this looks ok.

jimaek:
Can you minify the file by removing the comments as well? Make sure you do safe minification

megawac:
> Can you minify the file by removing the comments as well?

The comment looks fine to me, just tells the user where the code is from. (Weren't we leaning towards requiring that comment too, its an issue on bot somewhere)

akkumar:
Used this minifier - https://github.com/mishoo/UglifyJS2 to minify again. Please verify and let me know if this looks ok.

akkumar:
FYI - Per the feedback above, the comments have been removed now (with a better minifier). Thanks !

jimaek:
@jsdelivrbot squash and merge

PR #5106, opened 2015-06-03
lodash 3.9.3

akkumar:


akkumar:
Ok - So I am only adding a new version of lodash here , as needed by bower_components of webcomponents.

I am not changing the info.ini of lodash as part of this commit.

The message reported is - "hostname 'lodash.com' doesn't match either of '*.fronteers.nl', 'fronteers.nl'" .

Can someone suggest what needs to be done here ? Thanks !

jimaek:
Its a problem with their website and the bot cant load it. You can ignore it.
Thanks

PR #5159, opened 2015-06-04
added version 0.5.1 to angular.wizard

jacobscarter:


jacobscarter:
.less file is there because it is in previous versions and your docs say to mimic folder/file structure

jimaek:
You can enable auto-updates btw https://github.com/jsdelivr/jsdelivr#auto-updating

PR #5174, opened 2015-06-05
Updated to version 1.4.0

sancospi:
Updated to version 1.4.0. This new version includes a new file called "angular-message-format.min.js"


sancospi:
- It's talking bout a version I didn't update
- It's the official's angular minified file
- They are a new modules in angular 1.4.0. "You can write localizable plural and gender based messages in Angular interpolation expressions and $interpolate calls"

PR #5191, opened 2015-06-06
Pull Request for Korean-Webfonts

korusdipl:
Hi

It's been a while since I last posted the pull request.

As two months have passed and there was no specific claims from the foundries or bug reports from the users, I believe it is the time for an update.

There won't be any kind of this update in a year or so.

Thanks


korusdipl:
The main blank css was mistakenly dropped.
If it makes problem, let me know.

jimaek:
The mainfile must be present every single version, otherwise part of our functionality will break

korusdipl:
OK. I will repost with the css shortly.

PR #5195, opened 2015-06-06
korean-webfonts Update - Fixed

korusdipl:
Fixed pull req from https://github.com/jsdelivr/jsdelivr/pull/5191

Hope this resolves all issues


jimaek:
Why are you deleting some of the old files?

korusdipl:
Because the files deleted were invalid or loaded inappropriately with some browsers. Some of the files added are so-called 'patched' ones.

jimaek:
Is this PR still accurate? Are you sure that no websites will break if we remove some of the old files?
If someone is using them they will get 404 errors

korusdipl:
They won't feel much difficulty as I have mentioned in the official repo that the font files were still in beta and could be changed at any moment, and I've mentioned this commit as the first with all checks done and ready for production. Most users are probably known about this.

PR #5206, opened 2015-06-07
Added phoneformat.js

ewino:


files/phoneformat.js/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "phoneformat.js",
+  "repo": "albeebe/phoneformat.js",
+  "files": {
+    "basePath": "dist",
+    "include": ["phone-format.js", "phone-format.min.js", "phone-format.min.js.map"]
```
jimaek: Instead of that simply use `*.min.*`

ewino: Won't the bot reject the add because of too many commits? (the docs say
it's supposed to be exactly two commits - one for each file)

jimaek: Yes but the files need to be correct. You can always squash your commits or open a new PR.
In this case I can squash it for you.

ewino: I'm sorry, but I'm not fluent in git, and really lost my way through rebasing, merging and pushing stuff between several local and remote branches, without any idea how the pull request is affected by it (I've used the github web-interface for this PR).
Isn't there any other way to input these parameters to the download-bot?

megawac: @jimaek I thought hosting `.map` was all fine and good

jimaek: @megawac .map yes, but phone-format.js not. `*.min.*` will include map files as well

jimaek:
@jsdelivrbot squash and merge

files/phoneformat.js/update.json
megawac: Yes, but the `.map` file points to the unminified file (so will cause 404s)

https://github.com/albeebe/phoneformat.js/blob/master/dist/phone-format.min.js.map#L1

PR #5322, opened 2015-06-16
Added pathforajs

dylo0:


pnommensen:
@dylo0  thanks for contributing :)

PR #5335, opened 2015-06-17
Branch 1.4.x updated to 1.4.1

sancospi:
Branch 1.4.x updated to 1.4.1


sancospi:
No commit problems

megawac:
Thanks @sancospi, currently reviewing this.

tomByrer:
I kinda wish `angular-csp.css` included the header-info comment like the .js files do, but it is the same as source.  (& wish it was minified, but that is only a few bytes).

Also, seems `angular-mocks.js` is not really for production; not only it is not minified when all the other JS in the folder is minified, but seems it is [for testing](http://stackoverflow.com/questions/13465773/loading-mocks-into-an-angularjs-unit-test).  It is included in older versions, though perhaps by mistake.

megawac:
I think it's fine in angulars case. Could be useful for jsfiddle like
experiments

tomByrer:
> Could be useful for jsfiddle like experiments

Thought it would be good at first, but changed my mind; they could directly link to `https://code.angularjs.org/1.4.1/angular-mocks.js`.  I'd like to keep jsDelivr production-only, but obviously I don't decide such things :)

PR #5373, opened 2015-06-20
Added jquery.dirtyforms

NightOwl888:


files/jquery.dirtyforms/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "jquery.dirtyforms",
+  "repo": "NightOwl888/jquery.dirtyforms.dist",
+  "files": {
+    "include": ["jquery.dirtyforms.min.*", "plugins/*.min.*"]
```
megawac: Hey, do you not want to include the `.map` files? I would be fine with you just including all the `js` and `map` files in your repo

NightOwl888: Since all of the `.map` files also include `.min` in the name, this should work. Or am I mistaken in this assumption? 

My intention is also to put all of the plugins into the root `jquery.dirtyforms` directory (not in a `plugins` subdirectory). Please double check that will work too.

Note that it would be useful if you had a tool that would identify what files would be pulled by a specific `update.json` file and what their destination paths would be so we could make corrections before submission.

PR #5374, opened 2015-06-20
Create update.json for KaTeX

as-com:
I hope this works...

### DIFF
--- a/files/katex/update.json
+++ b/files/katex/update.json
@@ -1,0 +1,9 @@
+{
+  "packageManager": "bower",
+  "name": "katex",
+  "repo": "Khan/KaTeX",
+  "files": {
+    "basePath": "bower_components/katex/dist",
+    "exclude": ["README.md"]
+  }
+}


files/katex/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "bower",
+  "name": "katex",
+  "repo": "Khan/KaTeX",
+  "files": {
+    "basePath": "bower_components/katex/dist",
+    "exclude": ["README.md"]
```
megawac: I don't think this is right:

https://github.com/Khan/KaTeX/tree/v0.4.0

I think `/dist` would work though?

as-com: @megawac Ok, fixed that

files/katex/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "bower",
```
megawac: Whoops, I missed you were using `bower`. Can you let me know if its good to merge @as-com?

as-com: @megawac So do I need to add `/bower_components/katex`?

megawac: Or change to github. I'm not sure what the `bower` file structure looks like. I haven't used KaTex myself so I don't know what files should be expected

as-com:
Ok, I think it's ready to merge. I squashed the commits as well.

megawac:
Awesome thanks!

PR #5395, opened 2015-06-22
ng file upload

jbarrus:
ng file upload is already on jsdelivr but it's pointing to the wrong place and doesn't have update.json (so it's outdated) and isn't named consistently with the project.  There is a separate repo maintained for bower releases that should be used instead.

https://github.com/jsdelivr/jsdelivr/tree/master/files/angular.file-upload

### DIFF
--- a/files/ng-file-upload/info.ini
+++ b/files/ng-file-upload/info.ini
@@ -1,0 +1,5 @@
+author = "Danial Farid"
+github = "https://github.com/danialfarid/ng-file-upload-bower"
+homepage = "https://github.com/danialfarid/ng-file-upload"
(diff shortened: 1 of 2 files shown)


megawac:
LGTM

jimaek:
If we already host and then we should continue to use the old one instead of duplicating the projects. You can add the update.json to the old project and you can fix any broken URLs. 
Regarding the name its consistent with our naming  https://github.com/jsdelivr/jsdelivr/blob/master/CONTRIBUTING.md
>If the project is a plugin of a library, prepend the name of the library, like jquery.blurjs or bootstrap.select.

PR #5402, opened 2015-06-22
Added code-highlighter framework

avielfedida:


megawac:
@avielfedida you don't need to close, you can just add another commit to change the file

avielfedida:
@megawac I just figure it out, thank you, ill do so.

PR #5444, opened 2015-06-24
Fix mainfile for jsTree

as-com:


jimaek:
You will also need to add a mainfile overide to pre1.0 where the old format is located. Example https://github.com/jsdelivr/jsdelivr/blob/master/files/sweetalert/0.1.2/mainfile

as-com:
Alright, fixed that. The bot still complains though.

jimaek:
Yeah the bot is not smart enough yet :)

jimaek:
@jsdelivrbot squash and merge

PR #5453, opened 2015-06-24
Add update.json for when

bdukes:


files/when/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "bower",
+  "name": "when",
+  "repo": "cujojs/when",
+  "files": {
+    "include": ["when.js"]
```
megawac: That file won't work as it's importing a bunch of stuff https://github.com/cujojs/when/blob/master/when.js

bdukes:
yeah, it doesn't look like there's a minified version available through bower

PR #5507, opened 2015-06-27
Add js-cookie (former jquery.cookie)

FagnerMartinsBrack:
jQuery dependency was made optional in the version `1.5.0`. `js-cookie` repository contains [all `jquery-cookie` releases/tags](https://github.com/js-cookie/js-cookie/releases) since v1.

### DIFF
--- a/files/js-cookie/update.json
+++ b/files/js-cookie/update.json
@@ -1,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "js-cookie",
+  "repo": "js-cookie/js-cookie",
+  "files": {
+    "include": ["src/js.cookie.js"]
+  }
+}
--- a/files/js-cookie/info.ini
+++ b/files/js-cookie/info.ini
@@ -1,0 +1,4 @@
+github = "https://github.com/js-cookie/js-cookie"
+homepage = "https://github.com/js-cookie/js-cookie"
+description = "A simple, lightweight JavaScript API for handling cookies"
+mainfile = "js.cookie.min.js"
--- a/files/js-cookie/info.ini
+++ b/files/js-cookie/info.ini
@@ -1,3 +1,4 @@
+author = "js-cookie"
 github = "https://github.com/js-cookie/js-cookie"
 homepage = "https://github.com/js-cookie/js-cookie"
 description = "A simple, lightweight JavaScript API for handling cookies"
--- a/files/js-cookie/js-cookie
+++ b/files/js-cookie/js-cookie
@@ -1,0 +1,4 @@
+github = "https://github.com/js-cookie/js-cookie"
+homepage = "https://github.com/js-cookie/js-cookie"
+description = "A simple, lightweight JavaScript API for handling cookies"
+mainfile = "js.cookie.min.js"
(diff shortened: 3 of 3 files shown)


files/js-cookie/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "js-cookie",
+  "repo": "js-cookie/js-cookie",
+  "files": {
+    "include": ["src/js.cookie.js"]
```
jimaek: It needs to be `js.cookie.min.js` like you set in the mainfile.

FagnerMartinsBrack: What if we do not provide a minified file in the source repository? See [src/js.cookie.js](https://github.com/js-cookie/js-cookie/blob/master/src/js.cookie.js), the minified file is generated at build time.

The docs says:

> array that specifies files and dirs that will be copied from **project package**.

In this case the **project package** is the github, right?

jimaek: Do you deploy the min file to bower or npm? We can pull it from there as well

FagnerMartinsBrack: Neither, only the source. The client is responsible to minify the main file using their specific build routine.

For those using it in a simple website, they use CDN or online tools.

jimaek: Is there any chance you could add the min file somewhere? The CDN is all about the speed, but serving an unminified file defeats that purpose.

jimaek: Or if you dont want you can manually add the minified files in this PR. But then you will have to remove update.json because the bot wont be able to auto-update it.

FagnerMartinsBrack: Since the github repository is only supposed to provide source files (not production) the only way to handle this is having a release script to publish  the dist files for each tag. Unfortunately we do not have a release script yet.

I thought jsDelivr could handle unminified files, because cdnjs [also](https://github.com/cdnjs/cdnjs/pull/4713) [does it](https://github.com/cdnjs/cdnjs/commit/4d175c615fa6f123115b15118187c66e21ee39a3), but if there's no solution I think we may pass jsDelivr for now until we have a proper release script or when jsDelivr support file minification from source.

FagnerMartinsBrack: > Is there any chance you could add the min file somewhere?

The minified file is available as an attachment on each github release ([example](https://github.com/js-cookie/js-cookie/releases/tag/v2.0.2)).

jimaek:
If you agree we can add your project but it wont be auto-updated. And in the future when you have the release script we can enable auto-updates.

FagnerMartinsBrack:
I don't think we will have a release script very soon.

Is there any chances that jsDelivr implements a way to run the build script in the autoupdate process or a way to minify unminified files?

jimaek:
Its not planned. We want to avoid issues like bad minification resulting breaking thousands of websites. Thats why for automation we want to rely on the author to provide the correct file. Otherwise manually reviewed files for each new version.

FagnerMartinsBrack:
I understand. Unfortunately I am not able to manually send a PR for each released version, I will wait until someone requests jsDelivr for js-cookie or when we get a proper release process.

Thanks!

FagnerMartinsBrack:
I received a notification that js-cookie is on "jsDelivr" now. @jimaek are you ok for providing an unminified version? If I knew it was ok I would have kept this open -.-

jameswilson:
I'm curious, too. ^  the pull request (#7723) seems to only contain an unminified version:

https://github.com/jsdelivr/jsdelivr/pull/7723/commits/4859d3206ba8e6f5dd946e7af6a1f5d6f34cdfeb

PR #5523, opened 2015-06-28
Include example styles from colorbox

ntd:


jimaek:
Why do you want the examples? I dont think they are needed in a production environment

ntd:
They are really 5 different themes, with CSS and images (I purposely skipped the HTML files and the sample images). Colorbox without theme is quite ugly.

Unfortunately I did not find a way to rename that `example` prefix to something more meaningful such as `theme`.

ntd:
I just found this [uservoice post](http://jsdelivr.uservoice.com/forums/162344-general-ideas-and-feedback/suggestions/3630136-hosted-all-css-of-plugin-as-well) talking about these very same files.

If you have already planned to support renaming in libgrabber ([issue 10](https://github.com/jsdelivr/libgrabber/issues/10) is somewhat related) I can for now disable automatic update (e.g. rename `update.json` to `update.jsoff`) and update by hand using `theme1` instead of `example1`, `theme2` instead of `example2` and so on.

Just let me know.

jimaek:
Sorry I just dont like the idea of this. Its called `examples` and it hasnt been updated in 1 year. It doesnt look like part of production. Its like uploading old demo files

ntd:
No problems, I'll see if a directory refactoring on the colorbox side is feasible, in which case I'll open a new PR.

ntd:
In jackmoore/colorbox#733 I proposed a backward compatible workaround on the colorbox side. Let's see if we can reach an agreement.

PR #5543, opened 2015-06-29
Added version 4.1 of Syncano JS library

adamwardecki:


jimaek:
What is the source for this? The latest here is 3.x https://github.com/Syncano/syncano-js/blob/master/dist/syncano.min.js

adamwardecki:
Source for v 4.1 is in a different repo: https://github.com/Syncano/syncano4-js/tree/master/lib

PR #5556, opened 2015-06-30
Create info.ini

gustoase:


drozdov-ai-phone-call/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Alexandr Drozdov"
```
megawac: Place this under the `files/` folder

gustoase: hi) I added all the files

jimaek:
They are still not under /files

jimaek:
Also your project lack any documentation, we require that

PR #5560, opened 2015-07-01
Updated repository location, added latest release

bradberger:


bradberger:
@jimaek Sorry for the trouble, but I don't quite understand this error. Can you point me in the right direction?

jimaek:
Looks like SSL failed. I will look into it.
Why did you delete some of the files? We dont allow the removal of existing files

bradberger:
Ah, my bad regarding the removal. It's a branch that's between 0.3 and 0.5 and isn't used anymore, but keeping it there I guess makes sense for compatibility. I'll put it back and submit a new PR.

PR #5643, opened 2015-07-05
Fixed Encoding to UTF-8

sms200207:


sms200207:
It had problems in encoding.
I re-encoded them

sms200207:
@jimaek @megawac

megawac:
Can you remove the files and just set up auto updating (pointing at https://github.com/markquery/spidersweb/tree/master/images)? Thanks

PR #5655, opened 2015-07-06
highcharts ng initial version

akkumar:
Initial Version of highcharts-ng : 

Available at: https://github.com/pablojim/highcharts-ng


jimaek:
Please rename the directory to `angular.highcharts`.
The rest looks good

akkumar:
There seems to be another different project as well ( called https://github.com/rootux/angular-highcharts-directive ) , independent of this.  Hence, using the same name as the original author had used. 

Let me know if this looks ok.

jimaek:
For that project we could use `angular.highcharts-directive`

akkumar:
Renamed the package as per the discussion.

PR #5742, opened 2015-07-10
Adding Library Jquery.UI.Rotatable

abishekrsrikaanth:
Repo: https://github.com/godswearhats/jquery-ui-rotatable
Desc:  Adds a small handle to an HTML element to allow it to be rotated


jimaek:
:+1:

PR #5770, opened 2015-07-10
Add missing versions of angular

megawac:
Adds some versions missing from https://github.com/angular/code.angularjs.org


megawac:
Adds
```
files/angularjs/0.10.6/
        files/angularjs/1.2.28/
        files/angularjs/1.3.17/
        files/angularjs/1.4.2/
        files/angularjs/2.0.0-alpha.30/

```

PR #5786, opened 2015-07-11
Update project dot to 1.0.3

jsdelivrbot:
libgrabber updated project dot to 1.0.3.

@olado

### Metadata information

```json
{
  "metadata": {
    "packageManager": "npm",
    "name": "dot",
    "repo": "olado/doT",
    "files": {
      "include": ["doT.min.js", "doT.js"],
      "exclude": []
    },
    "path": "/home/jsdelivr/files/dot.js",
    "localVersions": [],
    "remoteVersions": ["0.1.0", "0.1.1", "0.1.2", "0.1.3", "0.1.4", "0.1.5", "0.1.6", "0.1.7", "0.2.0", "0.2.1", "0.2.2", "0.2.3", "0.2.4", "0.2.5", "0.2.6", "1.0.0", "1.0.1", "1.0.2", "1.0.3"],
    "branchVersions": []
  },
  "updateRemote": true,
  "metadataPath": "/home/jsdelivr/files/dot.js/update.json",
  "version": "1.0.3",
  "updated": true,
  "updatePath": "/home/jsdelivr/files/dot.js/1.0.3",
  "branch": "dot/1.0.3"
}
```

PR #5787, opened 2015-07-11
doT.js Author length fix

Jerskouille:
Here to fix the issue with the bot in  #5784 and #5786 .


megawac:
Thanks, its not going to be an issue in the future I think but appreciate it

PR #5792, opened 2015-07-11
fix project `name` in description

tomByrer:
I assume you want the jsDelivr folder name here?
(Should be accurate for future bots.)


jimaek:
Sorry, but what is the point of this fix?

pnommensen:
@tomByrer the proper convention is for the "name" field to be the name of the GitHub repo. The name of the repo is "moment-timezone" and therefore this pull request would be invalid.

tomByrer:
ah, cheers

PR #5801, opened 2015-07-11
Adds angular v2

megawac:


delamitry:
This commit broke the angularjs 'latest version' cdn urls...
Imo, angular2 files should be in a separate project directory.

megawac:
`/latest` should be expected to break. It would be safer to use `/1/` instead of `/latest`

tomByrer:
> /latest should be expected to break.

This is true, though I think our Readme should be updated to reflect this?

I recommend usually only grabbing the newest point version (in Angular's case, `/1.4/`).  Same recommendation with jQuery; they actively had supported a 1.x & 2.x branch at same time.  jQuery & Angular in particular, there have been many changes including APIs between 'minor' point versions.

That said, I still don't recommend hosting beta versions, let alone ALPHA versions that are updated every 10 days.  Angular in particular; people can use Google's CDN for alpha/beta which works *almost* as good as jsDelivr ;)
Even if jsDelivr can handle the harddrive & bandwidth for alpha/beta, IMHO better to conserve the man-hours.

So, my vote is -1

jimaek:
@megawac Sorry but this was a huge update that could impact lots of websites, thats why I reverted it. Please dont push this kind of updates without talking to me first.
Its not even production ready yet, its still in alpha, there is no point hosting it at this moment.

megawac:
All right, noted and won't do again in the future

tomByrer:
I believe it is great that you want the best for users though @megawac !

PR #5807, opened 2015-07-11
ace v1.2.0

coderaiser:
Added [ace v1.2.0](https://github.com/ajaxorg/ace-builds/releases/tag/v1.2.0)


megawac:
Hey, would you mind setting up auto updating for `ace`?

coderaiser:
It would be great, but folders named: `src-min` and `src-min-noconflic' and should be auto-renamed to `min` and `noconflict` on every release (because someone who first uploaded `ace` folders called them this way), or renamed once in `ace` repo.
I think we should ask @nightwing about it.

megawac:
Okay, file renaming is something I've wanted to add to the bot for a good while as well

PR #5815, opened 2015-07-12
Please add new project unify

mike-feldmeier:


jimaek:
You have no tags hers https://github.com/mike-feldmeier/unify/releases
The bot will fail

files/jquery.unify/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "unify",
+  "repo": "mike-feldmeier/unify",
+  "files": {
+    "include": ["unify.js", "unify.min.js"]
```
jimaek: Please remove `unify.js` from here. The min file is enough

mike-feldmeier:
All set with release tag and update file.  As a note, you may want to update the example to exclude the non-minified version if that is not desired.

PR #5844, opened 2015-07-13
merge cs-widget

jacob-israel-turner:


jimaek:
You have no tags here https://github.com/SpotTech/CS-Widget/releases so the bot will fail.

Also the code is so simple that I dont even see why would you want to make it a separate js. just tell people to add `<a href='http://www.cancerspot.co'></a>` wherever they want

jacob-israel-turner:
@jimaek Thanks, I'll create a tag.  I want to create a link to my script so I can update the widget when I want.  We're launching in the App store in a few weeks, and I want to be able to redirect traffic there instead of to cancerspot.co, without having to have each of our promoters update their code.  If you have a better suggestion, I'm all ears!

jacob-israel-turner:
This is also very preliminary.  The widget will likely become slightly more complicated before we actually use it. Updated image, etc.

PR #5902, opened 2015-07-16
Add jQuery.countdown

l4u:


files/jQuery.countdown/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Edson Hilios"
```
megawac: Change the folder name to lowercase

l4u: Updated to lowercase

PR #6025, opened 2015-07-21
Create backbone.syphon

megawac:


files/backbone.syphon/update.json
```diff
@@ -0,0 +1,12 @@
+
+    {
+        "packageManager": "github",
+        "name": "backbone.syphon",
+        "repo": ["marionettejs", "backbone.syphon"],
```
jimaek: this is broken

PR #6027, opened 2015-07-21
Create backbone.babysitter

megawac:


files/backbone.babysitter/update.json
```diff
@@ -0,0 +1,10 @@
+{
+    "packageManager": "github",
+    "name": "backbone.babysitter",
+    "repo": ["marionettejs", "backbone.babysitter"],
```
jimaek: broken

PR #6034, opened 2015-07-21
Create backbone.syphon

megawac:


jimaek:
Looks good, you could add a rule to exclude `backbone.syphon.js` as well

jimaek:
@jsdelivrbot squash and merge

PR #6038, opened 2015-07-21
Create backbone-metal

megawac:


jimaek:
whats wrong with the bot?

megawac:
Looks like it got an error page from the github api

jimaek:
Ok, then the same question as here https://github.com/jsdelivr/jsdelivr/pull/6034

jimaek:
https://github.com/jsdelivr/jsdelivr/blob/master/CONTRIBUTING.md
>If the project is a plugin of a library, prepend the name of the library, like jquery.blurjs or bootstrap.select.

So it would be `backbone.metal`

PR #6045, opened 2015-07-22
Move jssor.slider to /files

dan-sceene:
Accidentally committed jssor.slider to the root directory. Moving to /files.


megawac:
Whoops, missed that

dan-sceene:
@megawac, the bot never pulled in the jssor.slider files. Was that because it was originally merged into the root directory?

megawac:
`basePath` should be `js` it looks like

PR #6047, opened 2015-07-22
BlogWriter.css file update

Third9:
1. footer image tag effect add

2. image tag line align


megawac:
Can you drop the unminified files and v1.0, leaving only 1.1

Third9:
Sorry for the late response. As you said, i deleted files

PR #6115, opened 2015-07-25
fixed update.json (font-natosans-kr)

ajkj:
fixed update.json (font-natosans-kr)


jimaek:
The name is fine, why are you changing it?

ajkj:
is that name correct?
i thought name must match the github repo.

jimaek:
the name must be correct if you are using npm or bower. For github it can be anything

PR #6144, opened 2015-07-28
Create info.ini

julienetie:


files/request-frame/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Julien Etienne"
+github = "https://github.com/julienetie/request-frame"
+homepage = "https://www.npmjs.com/package/request-frame"
```
jimaek: You can use https://github.com/julienetie/request-frame as the homepage

jimaek:
You need the update.json file otherwise the bot wont know what to do

PR #6149, opened 2015-07-28
enable auto-update for gl-matrix

Swatinem:
The version they have on npm only includes the .min js file :-(

### DIFF
--- a/files/glmatrix/update.json
+++ b/files/glmatrix/update.json
@@ -1,0 +1,8 @@
+{
+  "packageManager": "npm",
+  "name": "gl-matrix",
+  "repo": "toji/gl-matrix",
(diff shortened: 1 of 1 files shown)


jimaek:
You can use github as source to add both files.
Remove the basepath and include both files using `include`
`src/gl-matrix.js`, `dist/gl-matrix-min.js`

Swatinem:
src/gl-matrix.js is a cjs module so that won’t work on a cdn.

I think I will have to bug the maintainers to just push a version bump to npm/github…

jimaek:
I would be cool in removing the unminified file if it was a major version change. But if we remove it now that could break some websites for people using aliased versions.

Issue #5796, opened 2015-07-11
Version ordering incorrect for esri.leaflet


patrickarlt:
Currently http://www.jsdelivr.com/#!leaflet.esri is showing 1.0.0-rc.8 as the latest release but the latest release version wise is 1.0.0.

```js
require('semver').gt('1.0.0', '1.0.0-rc.8'); // true
```

This is probably sorting alphabetically, sorting using `semver.gt` would solve this.

jimaek:
Its something that is going to be fixed in the new website

patrickarlt:
:+1:

PR #6202, opened 2015-07-31
Remove riot 2.2.2-beta to avoid the issue #5796

cognitom:
Thank you for the everything about jsdelivr! :)

If we can, we'd like to delete the beta release of riot.js, because we have the same issue #5796.


jimaek:
In most cases we dont allow the modification of existing files but in this case nothing should break thanks to version aliasing.

cognitom:
@jimaek Thank you so much!!

PR #6221, opened 2015-08-01
updated the source - v1.0 tag.

soundar24:


soundar24:
I want to add some additional changes in my version 1.0 files, so that i am again updated in the same version source. So can you please include these changes.

soundar24:
Can you please let me know what should i do now ?

jimaek:
Why are you doing that? Just release a new version with the new changes

soundar24:
Actually this is a small minor changes only.. for each changes creating a new version is not good.. and also i have some milestones.. when i complete each at that time only i planned to publish a new version.. And also recently only i published my plugin, still it does't reach with peoples, so that we can add this in the same version also..

soundar24:
@jimaek : here you have mentioned as merged, but still the changes are not updated in the cdn link. Can you please let me know the details ?

jimaek:
The old version was cached on the CDN.
I purged the cache for your files

soundar24:
@jimaek : Thank you so much, now it was updated in the CDN also.. :+1:

PR #6222, opened 2015-08-01
Add `dush` - event handling for browser and nodejs

tunnckoCore:


tunnckoCore:
@megawac is that okey, or im missing something?

files/dush/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "dush",
+  "repo": "tunnckoCore/dush",
+  "files": {
+    "include": ["dist/dush.js", "dist/dush.min.js", "dist/dush.standalone.js", "dist/dush.standalone.min.js"]
```
jimaek: Can you remove the unminified files? I dont think they are needed in production

tunnckoCore: for some reasons some times they are needed, imo

jimaek: angular.moment is there because of legacy reasons, the first version had it before we started pushing for min only https://github.com/jsdelivr/jsdelivr/tree/master/files/angular.moment/0.10.0

Are you 100% sure the non-min files are needed on the CDN? I can add them but only if they are really necessary

tunnckoCore: okey, i will in a bit. only from `update.json` or also from `files/1.0.0`?

tunnckoCore:
done

jimaek:
@jsdelivrbot squash and merge

PR #6320, opened 2015-08-09
Patch 1 for domcom

taijiweb:
# domcom 
domcom is a web framework to provide dom component

## Features
* the value of any property and other stuffs can be a function
The stuffs, like dom property, the test condition of If, Case component, the items of Repeat component, can be functions. Different from the usage of function as property value in jQuery, the functions are used to declares the computating method for every rendering.
* update only the difference between the dom and the cache
* skip the inative component when updating dom
* decouple with model and controller
* composable component
* the root of the component may be multiple dom elements

### DIFF
--- a/files/domcom/update.json
+++ b/files/domcom/update.json
@@ -4,6 +4,6 @@
   "repo": "taijiweb/domcom",
   "basePath": "dist/",
   "files": {
-    "include": ["domcom.js", "domcom.min.js"]
+    "include": [domcom.min.js"]
   }
 }
--- a/files/domcom/update.json
+++ b/files/domcom/update.json
@@ -4,6 +4,6 @@
   "repo": "taijiweb/domcom",
   "basePath": "dist/",
   "files": {
-    "include": [domcom.min.js"]
+    "include": ["domcom.min.js"]
   }
 }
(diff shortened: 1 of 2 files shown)


files/domcom/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "domcom",
+  "repo": "taijiweb/domcom",
+  "basePath": "dist/",
+  "files": {
+    "include": ["domcom.js", "domcom.min.js"]
```
jimaek: Can you remove `domcom.js`? I dont think its required in production

taijiweb:
@jimaek , thanks for your review. I've removed domcom.js, could you please to check it again?

jimaek:
@jsdelivrbot squash and merge

taijiweb:
@jimaek , I'm glad to see that my project was accepted, thank you very much. But I could not access the JS file from the cdn, anything else need to be done? I'm expecting your help. I tried the urls like below: http://cdn.jsdelivr.net/domcom/0.0.1/domcom.min.js
http://cdn.jsdelivr.net/domcom/domcom.min.js

jimaek:
`"basePath": "dist/",` needs to be inside `files` https://github.com/jsdelivr/libgrabber#examples

thats why it failed to update

PR #6322, opened 2015-08-09
fix wrong location of basePath in update.json

taijiweb:


jimaek:
1 commit is enough to change 1 file. Dont include the old ones

PR #6323, opened 2015-08-09
fix wrong location of basePath in update.json

taijiweb:


jimaek:
The bot will notify you when your project is live

PR #6328, opened 2015-08-09
rename metismenu project name

onokumus:


jimaek:
You have 3 unrelated commits here that create conflicts.

PR #6334, opened 2015-08-10
Add OpenLayers 3.8.2.

mstenta:


megawac:
Thanks

drupol:
@mstenta could you provide a update.json file so we can have the library auto updated ?
See: https://github.com/jsdelivr/jsdelivr/blob/master/README.md

mstenta:
@drupol - I'm not sure that's possible with the ol3 Github, because the ./build folder doesn't exist in the repo. Only in the packaged released. Unless that's what jsDelivr is looking at... not sure...

drupol:
Indeed, sorry for the mislead.

jimaek:
We also support npm/bower if you release the compiled files there

drupol:
Hi @jimaek, could you tell me more about that ?
Maybe @elemoine can help ?

drupol:
CDNJS does it I think already : https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/ol3/package.json

mstenta:
I think that file is the equivalent of info.ini in jsDelivr - it is for general library information - not for updating.

drupol:
mh mmh, we need to find a way to get Openlayers automatically updated on cdnjs and jsDelivr.

drupol:
Looks like ol3 is also available with npm, i'll check deeper tomorrow.

drupol:
Weird, ol3 is no more available :(

http://api.jsdelivr.com/v1/jsdelivr/libraries/ol3

mstenta:
Yes it is: http://cdn.jsdelivr.net/ol3/3.8.2/ol.js

Where did you get that URL from?

mstenta:
Wait - your URL works too now.

drupol:
Indeed, now it works... weird :(

jimaek:
@drupol You can find more info here regarding the auto-updates https://github.com/jsdelivr/libgrabber
Our bot can download new versions and files from github/bower/npm

PR #6370, opened 2015-08-12
angular directive for chartjs

akkumar:
Angular directive for chart.js .


akkumar:
The auto check seems to pass here. Let me know if there is anything I need to do on the same. Thanks !

jimaek:
Instead of adding the files how about adding an update.json file? https://github.com/jsdelivr/libgrabber

akkumar:
Thanks, my first stab at update.json here. Let me know if that looks ok.

jimaek:
@jsdelivrbot squash and merge

PR #6389, opened 2015-08-13
Add oneMapToRuleThemAll project

dpellier:


jimaek:
1. You dont have documentation
2. There are no versions tagged
3. Maybe its best to rename the folder to `onemaptorulethemall`?

I recommend developing the project further and when its ready leave a comment here

dpellier:
Sorry about the too soon PR.
Project has been fully documented, and a 1.0.0 version has been released.

For the name, I like to keep camelCase, is it a problem with your standards ?

jimaek:
Only lowercase. Also add an include parameter to use only `*.min.*` files

PR #6391, opened 2015-08-13
autocomplete for angular.js

akkumar:


akkumar:
angucomplete-alt package added.

PR #6395, opened 2015-08-13
Auto-update bootswatch

markcarver:
x-ref: #2515

PR #6399, opened 2015-08-13
[bootswatch] Remove explicit include and remove fonts from exclude

markcarver:
Follow-up to #6395. Since everything is getting excluded (that needs to be), it should be fine leave the includes at `**/*` actually.

This also fixed the fonts folder from getting removed.


jimaek:
The previous config was much better.
https://github.com/jsdelivr/jsdelivr/pull/6400

markcarver:
Indeed :-/

I should really be testing this locally, one sec (sorry)

PR #6396, opened 2015-08-13
Update project bootswatch to v3.3.5+2

jsdelivrbot:
libgrabber updated project bootswatch to v3.3.5+2.

@thomaspark

### Metadata information

```json
{
  "metadata": {
    "packageManager": "github",
    "name": "bootswatch",
    "repo": "thomaspark/bootswatch",
    "files": {
      "include": ["./*/bootstrap.css", "./*/bootstrap.min.css"],
      "exclude": ["*(2|api|assets|bower_components|custom|default|fonts|global|help|tests|.gitignore|.travis.yml|CNAME|Gemfile|Gruntfile.js|LICENSE|README.md|_config.yml|bower.json|composer.json|favicon.ico|index.html|package.json)"]
    },
    "path": "/home/jsdelivr/files/bootswatch",
    "localVersions": ["3.0.0", "3.0.3-b", "3.1.0", "3.1.1", "3.3.2"],
    "remoteVersions": ["v2.0.0", "v2.0.1", "v2.0.2", "v2.0.3", "v2.0.4", "v2.1.0", "v2.1.1", "v2.1.1+1", "v2.2.2", "v2.2.2+1", "v2.2.2+2", "v2.3.0", "v2.3.0+1", "v2.3.0+2", "v2.3.1", "v2.3.1+1", "v2.3.2", "v3.0.0-rc1", "v3.0.0-rc2", "v3.0.0", "v3.0.1", "v3.0.2+1", "v3.0.2+2", "v3.0.2", "v3.0.3+1", "v3.0.3", "v3.1.0", "v3.1.0+1", "v3.1.1", "v3.1.1+1", "v3.2.0+3", "v3.2.0+1", "v3.2.0+2", "v3.2.0+4", "v3.2.0", "v3.3.0", "v3.3.1", "v3.3.1+1", "v3.3.1+2", "v3.3.1+3", "v3.3.2", "v3.3.4", "v3.3.4+1", "v3.3.5+1", "v3.3.5", "v3.3.5+2"],
    "branchVersions": []
  },
  "updateRemote": true,
  "metadataPath": "/home/jsdelivr/files/bootswatch/update.json",
  "version": "v3.3.5+2",
  "updated": true,
  "updatePath": "/home/jsdelivr/files/bootswatch/3.3.5",
  "branch": "bootswatch/v3.3.5+2"
}
```

PR #6402, opened 2015-08-13
[bootswatch] Fix directory/file excludes

markcarver:
Ok, this should do it.

1st line: exclude directories. They must be globstar and globstar doesn't work with `*(a|b)`
2nd line: exclude additional files

x-ref: #2515, #6395, #6396, #6399, #6400


jimaek:
https://github.com/jsdelivr/jsdelivr/pull/6407

PR #6408, opened 2015-08-14
Patch 1

nmihajlovski:


jimaek:
I dont completely understand what it does. The project is related to Android. How a frontend CDN could help in any way?

nmihajlovski:
The project is a Web Framework for Java, and it is not related to Android. The presentation layer of the framework includes a pack of JavaScript libraries and extra functionality on top of them.

jimaek:
But are those js libraries supposed to work in the browser? Or are they supposed to work only locally?

nmihajlovski:
Yes, all of the JavaScript libraries from the pack are used in the browser, not on the server side. The pack includes some JQuery plugins, some Angular plugins, some other libraries like Underscore, Sweetalert, Select2, Dropzone, Typeahead.js, etc. Also, some framework-specific extra JavaScript utils are included in the pack. So, the CDN would be very useful for production use, because all the JavaScript + CSS dependencies for a web page would be provided as single JS and CSS files.

jimaek:
@jsdelivrbot squash and merge

PR #6411, opened 2015-08-14
Updating RequireJS libs

oller:
Unfortunately requireJs doesn’t keep a minified version under source
control, so manual updates it is!


oller:
This is a bit tricky here as the bot mustn't have applied these rules previously, I'm following the existing patterns for updates on this lib.

pnommensen:
@oller Yeah, I see what happened. 

You can remove the files the bot is yelling about and add the correct files for the version(s) you'd like to add. This message from the bot is mainly a FYI.

oller:
OK cool, have done @pnommensen 

This commit now just adds versions 2.1.17 and 2.1.20 removing the older versions I started adding in to try and appease the bot.

PR #6407, opened 2015-08-14
Update project bootswatch to v3.3.5+2

jsdelivrbot:
libgrabber updated project bootswatch to v3.3.5+2.

@thomaspark

### Metadata information

```json
{
  "metadata": {
    "packageManager": "github",
    "name": "bootswatch",
    "repo": "thomaspark/bootswatch",
    "files": {
      "exclude": ["2/**", "api/**", "assets/**", "bower_components/**", "custom/**", "default/**", "global/**", "help/**", "tests/**", "*(.gitignore|.travis.yml|CNAME|Gemfile|Gruntfile.js|LICENSE|README.md|_config.yml|bower.json|composer.json|favicon.ico|index.html|package.json)"],
      "include": ["**/*"]
    },
    "path": "/home/jsdelivr/files/bootswatch",
    "localVersions": ["3.0.0", "3.0.3-b", "3.1.0", "3.1.1", "3.3.2"],
    "remoteVersions": ["v2.0.0", "v2.0.1", "v2.0.2", "v2.0.3", "v2.0.4", "v2.1.0", "v2.1.1", "v2.1.1+1", "v2.2.2", "v2.2.2+1", "v2.2.2+2", "v2.3.0", "v2.3.0+1", "v2.3.0+2", "v2.3.1", "v2.3.1+1", "v2.3.2", "v3.0.0-rc1", "v3.0.0-rc2", "v3.0.0", "v3.0.1", "v3.0.2+1", "v3.0.2+2", "v3.0.2", "v3.0.3+1", "v3.0.3", "v3.1.0", "v3.1.0+1", "v3.1.1", "v3.1.1+1", "v3.2.0+3", "v3.2.0+1", "v3.2.0+2", "v3.2.0+4", "v3.2.0", "v3.3.0", "v3.3.1", "v3.3.1+1", "v3.3.1+2", "v3.3.1+3", "v3.3.2", "v3.3.4", "v3.3.4+1", "v3.3.5+1", "v3.3.5", "v3.3.5+2"],
    "branchVersions": []
  },
  "updateRemote": true,
  "metadataPath": "/home/jsdelivr/files/bootswatch/update.json",
  "version": "v3.3.5+2",
  "updated": true,
  "updatePath": "/home/jsdelivr/files/bootswatch/3.3.5",
  "branch": "bootswatch/v3.3.5+2"
}
```

PR #6415, opened 2015-08-14
[bootswatch] Finalize update.json

markcarver:
Also removes existing folders (as they were manually added and conflict with the update bot).

This is in response to https://github.com/jsdelivr/jsdelivr/pull/6407#issuecomment-131165033


jimaek:
Dont delete old folders, you will break websites that use those links. Just send the updated update.json

markcarver:
@jimaek, maybe, but the update isn't fixing the inconsistencies (that _should_ be fixed).

markcarver:
I've removed the last commit from this PR

PR #6422, opened 2015-08-14
[bootswatch] Fill in missing 3.x.x versions

markcarver:
Follow up to https://github.com/jsdelivr/jsdelivr/pull/6407#issuecomment-131165033


jimaek:
I understand that you want to have the same data in all versions. But I dont understand why.
You can ignore the useless files in the API and only use the ones that you are interested in

markcarver:
screen shot 2015-08-14 at 3 51 13 pm

This list is being populated from http://api.jsdelivr.com/v1/bootstrap/libraries

To "ignore" which one's aren't actually in https://github.com/jsdelivr/jsdelivr/tree/master/files/bootswatch, I would have to make an _additional_ HTTP request to http://api.jsdelivr.com/v1/jsdelivr/libraries/bootswatch to filter theme out.

I could also just use this API url specifically for bootstrap to begin with, but that is still 2 http requests. I'd rather not have to do that.

markcarver:
TY!

PR #6525, opened 2015-08-22
Create info.ini

Heark:


jimaek:
Your project is too fresh, doesn't explain what exactly it does and doesn't have any tagged versions.
Please re-submit in a few months when its more stable
Thanks

PR #6545, opened 2015-08-23
Add soma.js framework

Ragnarokkr:


files/soma.js/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "soma.js"
+github = "https://github.com/somajs/somajs"
+homepage = "http://somajs.github.com/somajs/"
+description = "soma.js is a scalable javascript framework created to help developers to write loosely-coupled applications to increase scalability and maintainability."
+mainfile = "build/soma-v2.1.1.min.js"
```
jimaek: The filename cant contain the version or else when a new version is release all links will break.

The correct would be `soma.min.js` but your repo doesnt have it

Ragnarokkr: I see. You're right. What if I do change the mainfile to the unminified version named `soma.js`?

jimaek: Technically you can but we really insist on using min files in production

Ragnarokkr: Yeah, that's true. Unfortunately, the author doesn't provides a minified version without the release hardcoded into the filename.

jimaek: The last release was a long time ago. I think we can disable the auto-updates and manually upload the latest version.
This way we will have the correct min file

Ragnarokkr: This would be a great solution, since it's not a bleeding edge library like other frameworks, such as Angular, Aurelia, etc. :+1:

jimaek:
Closed due to inactivity

PR #6572, opened 2015-08-25
Add js.js framework

sty1iss:
js.js is work like other javascript framework.
included default features like multi drop down navigation and slide content, ...

thanks for made this.


sty1iss:
did i something wrong?
this request seems to does not work.

jimaek:
No documentation, no stable versions and its too fresh.
Please develop it further and submit again in a few months when the project is more stable

sty1iss:
thanks for response.

PR #6578, opened 2015-08-26
Update jwplayer with 7.0.1

ping:
Build from https://github.com/jwplayer/jwplayer/releases/tag/v7.0.1


jimaek:
What is the source for these files? I've having trouble finding them on any official source

ping:
Like I mentioned above, it's at https://github.com/jwplayer/jwplayer/releases/tag/v7.0.1

There is no release build, so I downloaded the sources and ran a build using https://github.com/jwplayer/jwplayer/tree/v7.0.1#user-content-build-instructions

PR #6602, opened 2015-08-28
Change font-weight

ludens:


jimaek:
Why are you modifying existing files?

ludens:
NanumBarunGothic fonts have various font-weights.
But in the CSS files, there are just `font-weight:normal`. So I modified CSS files.

jimaek:
If you want to do modifications please create a new version. I dont want to break websites for people that already use these links.

PR #6635, opened 2015-08-31
Support source-map and non-minified version

garrypolley:
I'm updating this in hopes that we'll be able to support the minified, non-minified, and source maps for Canadarm.


megawac:
Sure thing, thanks

PR #6710, opened 2015-09-04
Update.json for socialite JS

isaackwan:
Do I need to send a PR to update the library to its latest version as well? Plus, am I expected to keep the unminified version in new versions?

### DIFF
--- a/files/socialite/update.json
+++ b/files/socialite/update.json
@@ -1,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "Socialite",
+  "repo": "tmort/Socialite",
+  "files": {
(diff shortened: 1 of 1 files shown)


files/socialite/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "Socialite",
+  "repo": "tmort/Socialite",
+  "files": {
+    "include": ["socialite.js>", "socialite.min.js", "extensions/*.js"]
```
jimaek: This will do what you want:
`"include": ["socialite.js", "socialite.min.js", "./extensions/*.js"]`

jimaek:
No need to add any other files

PR #6713, opened 2015-09-04
Auto-updating for adobe-webplatform/balance-text

redmunds:
Add update.json to setup auto-updating for adobe-webplatform/balance-text


jimaek:
Thank you :+1:

PR #6748, opened 2015-09-07
Add rtbid project

rtbid:


jimaek:
There is no documentation
The project is too fresh
There are no tags

Please develop it further and send a new PR again in a few months

PR #6769, opened 2015-09-08
leaflet.esri.renderers

jgravois:
thanks!

### DIFF
--- a/files/leaflet.esri.renderers/info.ini
+++ b/files/leaflet.esri.renderers/info.ini
@@ -1,0 +1,5 @@
+author = "Rachel Nehmer"
+github = "https://github.com/Esri/esri-leaflet-renderers.git"
+homepage = "https://esri.github.io/esri-leaflet"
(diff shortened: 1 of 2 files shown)


jgravois:
i'll drop `.git` from the url

files/leaflet.esri.renderers/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Rachel Nehmer"
+github = "https://github.com/Esri/esri-leaflet-related.git"
```
megawac: renderers not related

jgravois: fixed and rebased.  thanks!

someday i'll submit a PR here that doesn't have an immediately obvious problem :smile:

megawac: :D heh, I used to mess my submits all the time before I wrote a quick setup script https://github.com/megawac/cdn-it

PR #6814, opened 2015-09-11
Submitting RiotGear to the CDN

gregorypratt:


files/riotgear/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "npm",
+  "name": "riotgear",
+  "repo": "RiotGear/rg",
+  "files": {
+    "basePath": "dist/"
```
jimaek: Can you add an `include` parameter and include only the min file? The unminified file is not required in production

gregorypratt:
Updated to include an include parameter

jimaek:
@jsdelivrbot squash and merge

jimaek:
Thank you

PR #6827, opened 2015-09-11
add Leaflet.shapeMarkers

jgravois:
thanks (as always)!

cc/ @kneemer


jgravois:
i can rename to `leaflet.shapemarkers`

jimaek:
Please do that

jgravois:
closing in favor of #6830.  (darn mac isn't case sensitive enough to just fix the original PR)

PR #6868, opened 2015-09-14
react-quill

alexmojaki:


files/quill/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "quill",
+  "repo": "quilljs/quill",
+  "files": {
+    "include": ["*.js", "*.css"]
```
jimaek: Add a `basePath` parameter point to `dist/`

files/react.quill/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "react-quill",
+  "repo": "zenoamaro/react-quill",
+  "files": {
+    "basePath": "dist/"
```
jimaek: Add an `include` parameter with `*.min.*`

PR #6930, opened 2015-09-17
Update jwplayer with 7.1.0

ping:


jimaek:
what is 1.js?

ping:
I believe it is a polyfill script for IE.

PR #6956, opened 2015-09-17
Add typescript update.json

stevelacy:
Noticed that there isn't any auto update file for typescript, added it.
Please tell me if there are any issues.

### DIFF
--- a/files/typescript/update.json
+++ b/files/typescript/update.json
@@ -1,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "typescript.js",
+  "repo": "Microsoft/TypeScript",
+  "files": {
+    "basePath": "build/",
+    "include": ["*.js"]
+  }
+}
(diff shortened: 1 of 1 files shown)


files/typescript/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "typescript.js",
+  "repo": "Microsoft/TypeScript",
+  "files": {
+    "basePath": "build/",
```
jimaek: I dont see a build/ directory in the repo https://github.com/Microsoft/TypeScript/tree/v1.6.2

stevelacy:
@jimaek fixed

jimaek:
The files in lib/ dont look like production ready.
I think thats why we didnt enable auto-updates, there are no compiled files available

stevelacy:
Sounds like this pr can be closed then?

jimaek:
Maybe you could open a PR in their repo asking for the compiled browser files to be uploaded somewhere?

PR #6990, opened 2015-09-20
smart-sidebar

yaplas:
jquery plugin used to convert an element in a sidebar that follow the scroll in the right way, even when sidebar is higher than viewport.


megawac:
Hi, could you add a `git-tag` or `github release` to your repo. After that this looks good to go

jimaek:
Closed due to inactivity

PR #7002, opened 2015-09-21
Add covjson-reader

neothemachine:


files/covjson-reader/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "npm",
+  "name": "covjson-reader",
+  "repo": "Reading-eScience-Centre/covjson-reader",
+  "files": {
+    "include": ["*.js"]
```
jimaek: `*.min.js` is better for production

neothemachine: So only minified files should be included? This doesn't match the examples in the documentation. What about source maps? They need the unminified file as well, right?

jimaek: The ideal is to host only the minified file and nothing else. But its up to you if you want to add map files or anything else. I try to push for the min only in most cases

neothemachine: Ok, I changed it. No problem.

jimaek:
@jsdelivrbot squash and merge

PR #7020, opened 2015-09-22
get animate.css from bower

akkumar:
Get Animate.css from bower instead of github to be more authentic.


jimaek:
What do you mean more authentic? The repo is official, it looks good to me

PR #7065, opened 2015-09-24
add fancy-maps

osxi:


osxi:
Failing check says that that fancy-maps.min.js might not be properly minified. Ran it through `yuicompressor` and it looks/feels good to me.

jimaek:
if this is a plugin for react then rename the folder to `react.fancy-maps` please

osxi:
The project is not a React plugin but a library that wraps Leaflet.js *using* React with JSX.

Is it OK to place this PR on hold until I figure out a better way to build/minify the JS?

jimaek:
Sure. Leave a comment when you want to continue

osxi:
Thanks! It'll be sooner than later.

osxi:
@jimaek OK looks like the check passed this time! I'm ready.

PR #7074, opened 2015-09-25
Add cborjs-typedarray-decoder

neothemachine:
I would really like to include the .src.js and .min.js.map files. In my other package covjson-reader I was ok with skipping them as the .src.js file was already transpiled (from ES6) and it wasn't so useful in general. But here there's no transpilation so I find it very useful to include those files. Since this is a very small library I think it doesn't hurt. (both .src.js and .min.js.map are just 4k)

### DIFF
--- a/files/cborjs-typedarray-decoder/info.ini
+++ b/files/cborjs-typedarray-decoder/info.ini
@@ -1,0 +1,5 @@
+author = "Maik Riechert"
+homepage = "https://github.com/Reading-eScience-Centre/cborjs-typedarray-decoder"
+github = "https://github.com/Reading-eScience-Centre/cborjs-typedarray-decoder"
+description = A cbor-js decoder for CBOR typed arrays"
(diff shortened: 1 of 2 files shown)


jimaek:
@jsdelivrbot squash and merge

neothemachine:
Thanks, just had a look at https://www.jsdelivr.com/projects/cborjs-typedarray-decoder This seems broken. Any idea what's wrong?

jimaek:
@megawac @UnbounDev Any idea why libgrabber failed to add the files? When I run libgrabber manually I dont even see the error for this project. Its like the bot thinks it already has the latest version

neothemachine:
I just did a fake release on npm to try to trigger the bot. But I can't see any activity. Something is wrong here.

jimaek:
Yes. Not sure what. @megawac @UnbounDev  ping

neothemachine:
:cry:

PR #7115, opened 2015-09-28
updated Hack typeface webfonts to v2.015

chrissimpkins:


jimaek:
If you want you can enable auto-updates as well https://github.com/jsdelivr/jsdelivr#auto-updating

chrissimpkins:
Oh I will check that out.  Didn't know that.  That would be extremely helpful. As always thanks for this fantastic service!

PR #7174, opened 2015-10-01
Add update.json to /mousetrap

fatbattk:
For auto-updates.

### DIFF
--- a/files/mousetrap/update.json
+++ b/files/mousetrap/update.json
@@ -1,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "mousetrap",
+  "repo": "ccampbell/mousetrap",
+  "files": {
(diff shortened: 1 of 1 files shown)


files/mousetrap/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "mousetrap",
+  "repo": "ccampbell/mousetrap",
+  "files": {
+    "include": ["mousetrap.min.js", "mousetrap.js", "./plugins/**/*.js"],
```
jimaek: `"./plugins/**/*.min.js"` This will be better for production

fatbattk:
@jimaek Agree and updated. But why is auto-check failing for this?

    Issues validating update.json for mousetrap: No JSON object could be decoded

jimaek:
Not sure. Lets see what happens.

jimaek:
@jsdelivrbot squash and merge

PR #7201, opened 2015-10-03
add angular lib: ui-router-extras

christopherthielen:


files/angular.ui-router.extras/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "angular.ui-router.extras",
+  "repo": "christopherthielen/ui-router-extras",
+  "files": {
+    "basePath": "release/"
```
jimaek: Can you add an include parameter to make sure we only host min files?
`"include": ["*.min.*", "./modular/*.min.*"]`

jimaek:
Thank you

christopherthielen:
thanks!

PR #7223, opened 2015-10-05
Submitting clipboard.js 

gjmveloso:
Adding clipboard.js to jsDelivr. 
[Clipboard.js](http://zenorocha.github.io/clipboard.js/) is a lightweight and pure JavaScript library that allow us to copy text to clipboard without Flash dependencies

### DIFF
--- a/files/clipboard.js/update.json
+++ b/files/clipboard.js/update.json
@@ -1,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "clipboard.js",
+  "repo": "zenorocha/clipboard.js",
+  "files": {
+    "basePath": "dist/",
+    "include": ["*.js"]
+  }
+}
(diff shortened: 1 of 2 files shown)


files/clipboard.js/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "clipboard.js",
+  "repo": "zenorocha/clipboard.js",
+  "files": {
+    "basePath": "dist/",
+    "include": ["*.js"]
```
jimaek: How about using `*.min.*`? The unminified file is not needed in production

gjmveloso:
Thanks for the heads up @jimaek. Now autoupdate will just search for *.min.js files.

jimaek:
@jsdelivrbot squash and merge

PR #7259, opened 2015-10-06
Enable auto-update for Prism

Golmote:


files/prism/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "prismjs",
+  "repo": "PrismJS/prism",
+  "files": {
+    "include": ["prism.js", "components/**/*", "plugins/**/*", "themes/**/*"]
```
jimaek: This should work
`./components/*.min.js`
`./plugins/**/*.min.js`
`./themes/*.css`

Golmote: I've made the change.

jimaek:
@jsdelivrbot squash and merge

PR #7292, opened 2015-10-08
Added Angular Material 0.11.2

shumes:
I would like to add a new version of Angular Material


jimaek:
The files seem different from https://ajax.googleapis.com/ajax/libs/angular_material/0.11.2/angular-material.min.css and https://ajax.googleapis.com/ajax/libs/angular_material/0.11.2/angular-material.min.js
What is the source?

shumes:
The files were from Angular Material's Github page, https://github.com/angular/material. But I now realize we don't want those files. I've changed to the files from googleapis.com and squashed the commits. Hopefully this is ideal. Thanks!

jimaek:
:+1: thank you

PR #7331, opened 2015-10-12
Update matreshka.js structure

finom:


files/matreshka/update.json
```diff
@@ -3,6 +3,13 @@
   "name": "matreshka",
   "repo": "matreshkajs/matreshka",
   "files": {
-    "include": ["matreshka.min.js", "matreshka.js", "matreshka.min.map"]
+    "include": [
+		"./matreshka.min.js",
```
jimaek: Just `matreshka.min.js` here. `./` is only needed for folders where you need to keep the structure

finom: OK

PR #7387, opened 2015-10-14
Patch 3

labsvisual:


files/pigeon-js/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "pigeon-js",
+  "repo": "labsvisual/pigeon-js",
+  "files": {
+    "basePath": "dist/"
```
jimaek: Only `pigeon.min.js` is needed in production. Please add an include parameter for that file

labsvisual: Done!

jimaek:
@jsdelivrbot squash and merge

PR #7419, opened 2015-10-15
Update snowplow js to 2.5.2

apeckham:


jimaek:
If they publish the compiled min version somewhere you could enable auto-updates on this project https://github.com/jsdelivr/jsdelivr#auto-updating

PR #7438, opened 2015-10-18
update Vue.js repo

yyx990803:
The Vue.js repo just moved from `yyx990803/vue` to `vuejs/vue`. Is there anything else to make sure the auto update system picks up the change?


jimaek:
This should be enough. Thank you

PR #7761, opened 2015-11-05
updated Hack fonts to build v2.018

chrissimpkins:


jimaek:
You can enable auto-updates as well if you are interested https://github.com/jsdelivr/jsdelivr#auto-updating

PR #7779, opened 2015-11-06
Added files, added info.ini

hiccup01:
Added js, minified js and info.ini files.  Create structure to abide by jsDelivr rules.  http://phosphatejs.github.io/phosphateJS .


jimaek:
The project is too fresh. It has no tagged versions and no documentation.
I recommend developing it further and in the future if you are still interested adding it to jsDelivr open a new PR

PR #7807, opened 2015-11-07
Create info.ini + update.json for Honoka

yumeixxx:


files/honoka/update.json
```diff
@@ -0,0 +1,19 @@
+{
+  "packageManager": "github",
+  "name": "Honoka",
+  "repo": "windyakin/Honoka",
+  "files": {
+    "basePath": "dist/",
+    "include": [
+      "js/bootstrap.min.js",
```
jimaek: You need to add `./` before every path to make sure the path is retained

PR #7820, opened 2015-11-08
Semantic UI Daterangepicker

martykan:
Like the Bootstrap Daterangepicker, but for Semantic UI.

### DIFF
--- a/files/semantic-ui.daterangepicker/update.json
+++ b/files/semantic-ui.daterangepicker/update.json
@@ -1,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "semantic-ui.daterangepicker",
+  "repo": "BreadMaker/semantic-ui-daterangepicker",
+  "files": {
+    "include": ["daterangepicker.css", "daterangepicker.js", "daterangepicker.min.css", "daterangepicker.min.js"]
+  }
+}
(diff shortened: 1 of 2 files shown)


files/semantic-ui.daterangepicker/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "semantic-ui.daterangepicker",
+  "repo": "BreadMaker/semantic-ui-daterangepicker",
+  "files": {
+    "include": ["daterangepicker.css", "daterangepicker.js", "daterangepicker.min.css", "daterangepicker.min.js"]
```
jimaek: Most of these files are not here https://github.com/BreadMaker/semantic-ui-daterangepicker/tree/v1.3.21

martykan: The older versions use a different naming scheme.

jimaek: 1.3.21 is the latest version according to this https://github.com/BreadMaker/semantic-ui-daterangepicker/releases

martykan:
The releases seem to be of the Bootstrap version. I will close this request since the project needs to get the releases sorted out, before it can be published here.

PR #7888, opened 2015-11-12
Add prelodr v1.0.5

joseluisq:


files/prelodr/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "prelodr",
+  "repo": "quintana-dev/prelodr",
+  "files": {
```
jimaek: You need a basePath parameter  set to `dist/` since the files are here https://github.com/quintana-dev/prelodr/tree/1.0.5/dist

files/prelodr/update.json
```diff
@@ -0,0 +1,8 @@
...
+    "include": ["prelodr.js", "prelodr.min.js"]
```
jimaek: There is no need for `prelodr.js` the min is enough.

joseluisq: Ok @jimaek, I will remove `prelodr.js`.

files/prelodr/update.json
joseluisq: @jimaek ok, I will add `dist/` path.

joseluisq:
https://github.com/jsdelivr/jsdelivr/pull/7899

PR #7919, opened 2015-11-13
Add chassis

coreybutler:
Chassis is the UI component of a new & complete OSS microservice orchestration system.


megawac:
@jsdelivrbot squash and merge

PR #7932, opened 2015-11-14
Add wheelnav

softwaretailoring:


files/wheelnav/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "wheelnav",
+  "repo": "softwaretailoring/wheelnav",
+  "files": {
+    "include": ["wheelnav.js", "wheelnav.min.js", "wheelnav.min.js.map"]
```
jimaek: You need to add a `basePath` parameter to `js/dist` for this to work.
Also I dont think wheelnav.js is needed. The min file should be enough

softwaretailoring: Thank you for response!
I supposed that the robot reads the 'js/dist' path from package.json.
The update.json example contains the non-minified version, so I believed it is needed.

softwaretailoring: I take it already now, I found this page: https://github.com/jsdelivr/libgrabber#add-updatejson-schema
The update.json fixed.

jimaek:
@jsdelivrbot squash and merge

PR #7936, opened 2015-11-14
Update riot available files

GianlucaGuarini:
The main riot repo does not contain the `compiler.js` and `compiler.min.js` files anymore. We provide `riot+compiler.js` and `riot+compiler.min.js` instead


GianlucaGuarini:
thank you

PR #7959, opened 2015-11-16
Update update.json

unpete:


jimaek:
What are you trying to do? You want to include all the files under lib/ ? Why?
Most dont look like production files

unpete:
now, unnecessary files removed from the folder lib, saved only production files + images, used in css

unpete:
Please clarify that this is not so with files in the folder "lib/"? Saved only minified version.

PR #8004, opened 2015-11-19
initial version of base css framework

akkumar:
base css framework at: http://getbase.org/docs/ .


jimaek:
1. Your update.json is incorrect
2. The files here dont seen to be production ready https://github.com/matthewhartman/base/tree/master/dist

PR #8011, opened 2015-11-19
Adding Library Event.js

abishekrsrikaanth:
Repo: https://github.com/mark-rolich/Event.js

Repo does not include releases so have not added update.json. Added folder `0.1` as the release folder


jimaek:
This project looks very outdated, why do you need it?

PR #8032, opened 2015-11-20
Add tadaboard-embed

alex88:
I've added a library for a project we're working with, let me know if there is any problem, the library is open source but the project is a commercial one


jimaek:
The project lacks documentation and tagged versions. I recommend developing it further until its more stable and ready for production.

PR #8063, opened 2015-11-23
[author] Update stacktrace.js and add error-stack-parser

ssenbud:
This PR adds versions 0.6.4 and 1.0.1 of stacktrace.js. 

Version 1.0.1 of stacktrace.js split out Error parsing into a separate module, error-stack-parser, so we are adding that.

Thank you for hosting!


ssenbud:
Regarding the `update.json` here, is it possible that "includes" or "excludes" are required? The lodash example seemed to suggest that "baseDir" is necessary. Could someone advise here?

Regarding the file structure, these new files are expected and I hope that jsdelivr will consider hosting them because we'd really prefer hosting files that one can just "drop into" a webpage like the "-polyfill" variants. Please let us know if we really ought to remove these files.

jimaek:
@jsdelivrbot squash and merge

jimaek:
Everything looks good. Thank you

eriwen:
Late reply here: thank you @jimaek! Happy New Year!

PR #8070, opened 2015-11-23
Create update.json file for jquery.datatables

swcheon:


jimaek:
We cant auto-update this project because we also host the plugins for the previous version http://www.jsdelivr.com/projects/jquery.datatables

PR #8113, opened 2015-11-25
Added the new version (1.1) of roundSlider

soundar24:
Updated some fixes and added some improvement in roundSlider plugin. So published these changes as the next version (1.1) of roundSlider.

PR #8132, opened 2015-11-27
v1.1 - Change event related issue solved.

soundar24:
Need to update the v1.1 source, since in that i missed a major mistake. So need to include this commit also in this version, and it is necessary to reflect this changes in CDN also.


soundar24:
Hi there, Sorry for the inconvenience. Yesterday i created a pull request for v1.1 CDN and also it updated successfully. 
https://github.com/jsdelivr/jsdelivr/pull/8113

The problem is, in that version i missed a small (but major) correction.. So i need to update the mentioned commit also with that CDN.. So can you please merge these changes also.. Thanks...

megawac:
Hi @soundar24, you'd be better of creating a new version. Although modifying the source may seem more convenient/better, some users may already have 1.1 cached

soundar24:
@megawac : Actually still i didn't rollout this version, today only i planned to publish this.. but due to this correction i can't .. without this changes this version remains useless.. By skipping this version and release next version also no meaningful.. So can you please try to merge with this one.. Thanks...

megawac:
As I mentioned in the previous comment, this version is live on the CDN and
thus some users/proxies may have already cached the file. Modifying the
file may cause some users to have invalid files.

You'd be better off to publish a v1.1.1 or something

On Fri, Nov 27, 2015 at 12:52 PM, Soundara Pandian <notifications@github.com

soundar24:
yes @megawac , I can understand.. But still i didn't publish the new CDN.. and also no one know about this new version.. so definitely i can say no one didn't used this CDN still.. So that i request you to update in the same version..
If possible please update the needful changes..

PR #8168, opened 2015-11-30
Add Gridlex

devlint:


files/gridlex/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "Gridlex",
+  "repo": "devlint/gridlex",
+  "files": {
+    "include": ["dist/gridlex.min.css", "dist/gridlex.css"]
```
jimaek: I think the min file is enough for production

files/gridlex/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Laurent G."
+github = "https://github.com/devlint/gridlex"
+homepage = "http://gridlex.devlint.fr"
+description = "Gridlex is just a CSS Flexbox Grid System "
+mainfile = "dist/gridlex.min.js"
```
jimaek: 1. Where min.js came from? You only included min.css
2. Dont use `dist/`. Just `gridlex.min.js` is correct

files/gridlex/update.json
devlint: You're right, I've keeped only `gridlex.min.css`.

files/gridlex/info.ini
devlint: Again you're right, there is no JS files, I've made it too quickly.
For the dist/ it's because I've a `src` and a `dist` folders where the min.css file is. So for you I only need to target directly `gridlex.min.css` and not `dist/gridlex.min.css`?
Thank you!

jimaek:
@jsdelivrbot squash and merge

PR #8248, opened 2015-12-02
update.json references incorrect folder name

modethirteen:
Small naming oversight that caused the bot to fail


jimaek:
Everything is good https://github.com/jsdelivr/jsdelivr/pull/8246
its live

PR #8250, opened 2015-12-03
Ignore geniuslink.js and geniuslink.js.map

modethirteen:


jimaek:
@jsdelivrbot squash and merge

PR #8345, opened 2015-12-07
Update project jquery.hotkeys entirely valid version

swcheon:


swcheon:
Could you check this commits manually? :) Thanks.

jimaek:
You are changing the contents of an existing version. this is not allowed. You can only submit new versions

PR #8420, opened 2015-12-10
Add FileBrowser

jonataswalker:
Add FileBrowser

### DIFF
--- a/files/filebrowser/update.json
+++ b/files/filebrowser/update.json
@@ -1,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "FileBrowser",
+  "repo": "jonataswalker/FileBrowser",
+  "files": {
+    "basePath": "build/",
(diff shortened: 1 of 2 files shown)


files/filebrowser/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "FileBrowser",
+  "repo": "jonataswalker/FileBrowser",
+  "files": {
+    "basePath": "build/",
+    "include": ["filebrowser.js", "filebrowser.css", "filebrowser.min.css"]
```
jimaek: Why do you have both filebrowser.min.js and filebrowser.js in the repo? The both look to be minified

jonataswalker: It was an unneeded (old) file. Removed now.

jimaek:
@jsdelivrbot squash and merge

PR #8426, opened 2015-12-10
ls-chunk-store initital

DiegoRBaquero:
ls-chunk-store is a browser localStorage chunk store that is abstract-chunk-store compliant, usable by WebTorrent project and any project that uses abstract-chunk-store stores.

https://github.com/DiegoRBaquero/ls-chunk-store
https://www.npmjs.com/package/ls-chunk-store


jimaek:
@jsdelivrbot squash and merge

PR #8496, opened 2015-12-15
Add project foundation-sites with flex-grid enabled

andrewbiddinger:
Add project foundation-sites with flex-grid enabled, imported, compiled, and minified from scss. http://foundation.zurb.com/sites/docs/flex-grid.html


jimaek:
I was too fast to merge. Sorry but I cant add it. This project is auto-updated by out bot, adding this file manually would break the automation.

andrewbiddinger:
Oops. Sorry about that. Would it solve the problem to update the "update.json" file to include the "foundation.flex-grid.min.css" or does it need to be a new project?

jimaek:
That file is not hosted in their repo/npm. So adding it to update.json wouldnt do anything.
Isnt flex-grid part of the foundation 6? Why it needs to be a separate file?

andrewbiddinger:
@jimaek Flex-Grid (http://foundation.zurb.com/sites/docs/flex-grid.html) isn't enabled by default. In order to use it, you have to disable the default grid (scss) and enable flex-grid (scss) before the project is compiled to css. And you can't use flex unless this is done.

andrewbiddinger:
@jimaek Is this something that foundation would have to add as a sepperate file in their repo/npm?

jimaek:
If we want to have it under the `foundation` project then yes. We could create a new project for it but it wouldn't be auto-updated because there is no source and I dont want that.

andrewbiddinger:
Sounds good. I have put in that feature request with Zurb. Thanks for your help!

PR #8495, opened 2015-12-15
Update project scrollReveal to v3.0.0

jsdelivrbot:
libgrabber updated project scrollReveal to v3.0.0.

@jlmakes

### Metadata information

```json
{
  "metadata": {
    "packageManager": "github",
    "name": "scrollReveal",
    "repo": "jlmakes/scrollReveal.js",
    "files": {
      "basePath": "dist/",
      "include": ["**/*"],
      "exclude": []
    },
    "path": "/home/libgrabber/jsdelivr/files/scrollreveal.js",
    "localVersions": ["0.1.2", "2.3.2"],
    "remoteVersions": ["v2.0.0", "v2.0.1", "v2.0.2", "v2.0.3", "v2.0.4", "v2.0.5", "v2.1.0", "v2.2.0", "v2.3.0", "v2.3.1", "v2.3.2", "v3.0.0"],
    "branchVersions": []
  },
  "updateRemote": true,
  "metadataPath": "/home/libgrabber/jsdelivr/files/scrollreveal.js/update.json",
  "version": "v3.0.0",
  "updated": true,
  "updatePath": "/home/libgrabber/jsdelivr/files/scrollreveal.js/3.0.0",
  "branch": "scrollReveal/v3.0.0"
}
```

PR #8511, opened 2015-12-16
Update project scrollreveal.js to v3.0.0

jlmakes:
Everything looks nice and tidy for the new release. I also removed an ancient version of ScrollReveal that hasn’t been supported for over a year.


jlmakes:
:+1:

jimaek:
I am sorry but we dont allow the removal of existing files, if someone was using it it would break their site.

jimaek:
Actually I just merged https://github.com/jsdelivr/jsdelivr/pull/8495 since everything was ok. 
Future updates should just work just fine as well

jlmakes:
Thanks @jimaek, but the problem is that file case has changed, and the `mainfile` entry is now incorrect. It reads `scrollReveal.js` but should be `scrollreveal.js`. Also the library description was updated, and disregarded.

I understand not wanting to break people's sites, but now there's a 404 on the jsdeliver site...

jimaek:
It should be fixed now, sorry about that

jlmakes:
@jimaek cheers!

PR #8531, opened 2015-12-17
what-input

TalhaAzim:
Added info.ini and update.json for what-input script used by foundation 6.


jimaek:
@jsdelivrbot squash and merge

PR #8577, opened 2015-12-19
don't export sfx bundles

rosendi:


jimaek:
Why? We already host these files so this change would require a major version change to make sure we dont break any sites.

rosendi:
Well, the SFX-bundles are redundant and I won't provide them once 0.2.0 is landed. Having of the UMD-bundles is enough, since the library is used in the Angular 2 ecosystem and the recommended loader is SystemJS. The public api can be changed without a major version bumping until the 1.0.0 release.

jimaek:
If the next update will be 0.2.x then sure. But if you plan to release some 0.1.x versions then I cant merge this yet

rosendi:
Ah, OK! I thought that I have to update `update.json` before the upcoming release.

rosendi:
Yes, the next update is 0.2.0.

rosendi:
@jimaek thank you!

PR #8582, opened 2015-12-19
[new] redux-toastr

diegoddox:
The full project you can find it here https://github.com/diegoddox/redux-toastr


jimaek:
You dont explain anywhere in your documentation what your project does. You also dont have any tagged versions. Plus the project is very new, I would wait until it becomes more stable before submitting it to jsDelivr

PR #8675, opened 2015-12-28
Add font PT Sans.

iMega:


jimaek:
We already host it http://www.jsdelivr.com/projects/font-pt-sans

PR #8717, opened 2015-12-31
ScrollReveal Cleanup

jlmakes:
@jimaek I made the ideal changes I _think_—now you tell me, what is wrong with what I’m doing? Is there a way to see usage statistics to know how many users/sites would break by removing the inconsistently cased files from `2.3.2` ??


jimaek:
I am not sure what you are trying to do here. All you have to do is basically add the missing files that the bot complains about and thats it. Dont remove anything.
Also unfortunately there is no way to see the stats

PR #8763, opened 2016-01-03
Patch 1

icanget:
Released version 1.1 of javascript styles composer.


jimaek:
What exactly this project does? Does need to run every single time a user visits a page? It looks like a one time thing

icanget:
Hello. This project generates responsive styles using javascript at the user side. It generates common styles properties with their values in way that developer can assign them by classes and have them responsive by 4 type of devices (similar to Bootstrap logic). It has to be run every single time a user visits a page. Sure, those styles can be as a separate .css file, but generating them via Javascript saves loading time (file size of .js file is in 30 times less compared to generated .css file).

PR #8898, opened 2016-01-12
Add "kick_sovetnik"

SerjoPepper:


jimaek:
The project was created 20 hours ago. I would wait a while until it becomes more stable before publishing it here

PR #8929, opened 2016-01-13
update(update.json)

jlmakes:
Attempting to fix jsdelivr’s mad problems with my automatic PR merges… I don’t see a reason to include the non-minified version, since it seems to just be causing problems.


jimaek:
We cant remove that file to not break existing websites that use it.

PR #8944, opened 2016-01-14
Fix #8922

Kcchouette:
Change the mainfile to be the min version for leaflet.locatecontrol


Kcchouette:
If you can remove the v-0.19.0 of jsdelivr, thanks :)

PR #9024, opened 2016-01-17
Added two angular tools

JohnnyTheTank:


tomByrer:
We cleared things up in private chat; should be G2G.

jimaek:
@jsdelivrbot squash and merge

PR #9159, opened 2016-01-18
updated Hack web font release to v2.019

chrissimpkins:
Font files and CSS files updated for the v2.019 release


chrissimpkins:
The removal of the svg files was intentional.  They will no longer be included in our web font releases.

jimaek:
Thank you

chrissimpkins:
Thanks!

PR #9298, opened 2016-01-23
roundSlider - new version 1.2

soundar24:
Added the new version (1.2) of roundSlider.


soundar24:
Need to add the 1.2 folder alone for the new version. Can you include the new version in the CDN.

jimaek:
@jsdelivrbot squash and merge

soundar24:
@jimaek: I already delivered v1.1, now i updated to v1.2.. But i don't need the v1.1 since it have some bugs.. so is it possible to remove that v1.1 ?

jimaek:
No. We dont remove any files to avoid breaking websites that could use it.

soundar24:
However that is updated by me only, but due to the bug i wanted to remove that.. Also that is buggy, so no one can't use that (event using that is worthless)... also that was unwantedly occupied the space in the server.. 

Also keeping that files may cause confusion, so that i am thinking to remove.... 

Atleast here can i show the stable versions only ?  (in the below link the all versions are shown in the dropdown)
http://www.jsdelivr.com/projects/jquery.roundslider

jimaek:
If you want you can replace the 1.1 files with 1.2 files. This way there wont be any 404 errors

soundar24:
Even if I replace, will it be updated in the CDN too ?

jimaek:
Yes, it will take some time to update but yes. If you are in hurry I can manually purge the cache

soundar24:
Ooh thanks and great news... If you can manually purge the catch, then I will commit the particular code alone in the 1.2 version.. So the bug will be solved from the 1.2 version.. 

Is that possible ?

jimaek:
Yes

PR #9402, opened 2016-01-27
jsDelivr auto-update for SharepointPlus

Aymkdn:
Hi,

I want to try the auto-update process.

Thanks

### DIFF
--- a/files/sharepointplus/update.json
+++ b/files/sharepointplus/update.json
@@ -1,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "sharepointplus",
+  "repo": "aymkdn/sharepointplus",
+  "files": {
+    "basePath": "jsdelivr/"
+  }
+}


files/sharepointplus/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "sharepointplus",
+  "repo": "aymkdn/sharepointplus",
+  "files": {
+    "basePath": "jsdelivr/"
```
jimaek: There is no such folder here https://github.com/Aymkdn/SharepointPlus/tree/v3.12

Aymkdn: Sorry about it. It should be ok now!

jimaek: Since we already host the unminified file here http://www.jsdelivr.com/projects/sharepointplus we need to add it to updates as well. Unless you release a major version, then we dont worry about breaking the file structure

Aymkdn: No need for the unminified version anymore. And I decided to use a more
correct version number, it's why it's now v3.12

Aymkdn: Do I need to do anything else?

Thanks!

PR #9455, opened 2016-01-29
wdt-emoji-bundle repo

needim:
Slack like emoji selector with apple/ios, twitter/twemoji, google, emojione and custom emoji support
Add wdt-emoji-bundle https://github.com/needim/wdt-emoji-bundle


jimaek:
You are missing the update.json file that does the auto-updating. 
https://github.com/jsdelivr/jsdelivr#auto-updating

PR #9466, opened 2016-01-30
Add linkify.js to jsdelivr

karololszak:


karololszak:
Now this is ridiculous - documentation says "files/include (optional)", yet my pull request was denied because it was missing!

Anyhow, here is the repo of the project: https://github.com/SoapBox/linkifyjs/
It is available on github in a shim form, which I linked to, but it is also available on npm and bower so if somebody wants to fix this PR please do so.

jimaek:
include is optional, but `files` is not optional. You are missing that part completely

karololszak:
According to [this documentation](https://github.com/jsdelivr/libgrabber#add-updatejson-schema), `update.json` should look like this:
```
{
  "packageManager": "<github|npm|bower>",
  "name": "<package-name>",
  "repo": "<user>/<repo>",
  "files": {
    "basePath": "<dir>",
    "include": ["<glob-string-1>", "<glob-string-2>"],
    "exclude": ["<glob-string-3>"]
  }
}
```
where [...]:
- `files/basePath` (optional) [...]
- `files/include` (optional) [...]
- `files/exclude` (optional) [...]

So... what exactly should I put in `files`, if everything is optional? `"files": {}`?

jimaek:
1. If you really want to add everything then you will need to add `basePath: /`. But I dont think there are cases where you truly dont need a single parameter in `files`
2. In your case your config is completely wrong. You need to add the browser production files and not just all the source code

karololszak:
Ad. 1, What I tried to show is that if everything is optional, `files` shouldn't be required. If it is required, then maybe the docs/reference should be updated - at least to state that it can be left empty to pass the automatic checks (like I did to cheat the system).

Ad. 2, The browser production files are [here](https://github.com/nfrasser/linkify-shim) - and I think I referenced that "repo" in `update.json` correctly. [Source files are here.](https://github.com/SoapBox/linkifyjs/)
Linkify needs all of production files from the "shim", because usage examples look like this:
```html
<script src="linkify.js"></script>
<script src="linkify-string.js"></script>
<script>
var str = "Some string www.google.com more string...";
var linkified = str.linkify(); // "Some string <a href='http://www.google.com'>www.google.com</a> more string...";
</script>
```
So you can pick-n-choose what modules you need (and it would be great when paired with jsdelivr chaining of assets - the "load multiple files with single request" feature).

PR #9467, opened 2016-01-31
Adding sinon.js

RandomlyKnighted:
sinon.js creates new releases regularly. However, they only provide a zip file for each release. Is it possible to set up the project to automatically pull each release if that's all they provide or will we need to submit a PR each time a release is made?

### DIFF
--- a/files/sinonjs/info.ini
+++ b/files/sinonjs/info.ini
@@ -1,0 +1,5 @@
+author = "Maximilian Antoni"
+github = "https://github.com/sinonjs/sinon"
+homepage = "http://sinonjs.org/"
+description = "Sinon.js is a standalone and test framework agnostic JavaScript test spies, stubs and mocks."
+mainfile = "sinon.js"


files/sinonjs/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "npm",
+  "name": "sinon",
+  "repo": "sinonjs/sinon",
+  "files": {
+    "basePath": "lib/",
+    "include": ["sinon.js"]
```
jimaek: I dont think that file is ready to work in the browser. This is what we need http://sinonjs.org/releases/sinon-1.17.3.js

RandomlyKnighted: @jimaek Is it possible to set up auto updates on this project? I ask because as you said their NPM files don't look to be ready to setup in the browser. Also their GitHub releases don't have any files except a .zip file of the whole project and they don't release on Bower.

jimaek: As far as I see its not possible. The authors are not publishing the production files anywhere where our bot could take them. Sorry

RandomlyKnighted: If they publish the production files in a GitHub release will that be enough for the bot to pull them? I'm happy to submit a issue on their repo requesting this.

jimaek: The bot is using Github tags. For example https://github.com/sinonjs/sinon/tree/v2.0.0-pre
If they add their files there in all new versions then it will work.

RandomlyKnighted: Awesome. I'll submit the issue now. What should i do about this issue in the meantime?

PR #6433, opened 2015-08-16
Add jQuery-QueryBuilder

mistic100:
http://mistic100.github.io/jQuery-QueryBuilder

PR #9619, opened 2016-02-06
Add Photo-Sphere-Viewer

mistic100:
http://mistic100.github.io/Photo-Sphere-Viewer (page is broken waiting for dns.js.org approval)

### DIFF
--- a/files/photo-sphere-viewer/info.ini
+++ b/files/photo-sphere-viewer/info.ini
@@ -1,0 +1,5 @@
+author = "Damien 'Mistic' Sorel"
+github = "https://github.com/mistic100/Photo-Sphere-Viewer"
+homepage = "http://mistic100.github.io/Photo-Sphere-Viewer"
+description = "A JavaScript library to display Photo Sphere panoramas"
+mainfile = "photo-sphere-viewer.min.js"
--- a/files/photo-sphere-viewer/update.json
+++ b/files/photo-sphere-viewer/update.json
@@ -1,0 +1,9 @@
+{
+  "packageManager": "bower",
+  "name": "Photo-Sphere-Viewer",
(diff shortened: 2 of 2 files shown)


mistic100:
@jimaek I don't understand the check error :s

jimaek:
1. Why are you using a fork and not the original repo?
2. Why are you trying to add `./`? It makes no sense in this case

mistic100:
1. My fork diverged from the original project in many ways, the author didn't want to publish to Bower, I do.
2. I take my other repo as reference and your comment https://github.com/jsdelivr/jsdelivr/pull/6433#discussion_r37149962, re-reading the doc `*.min.*` should be enough

jimaek:
I recommend renaming the folder to something like `mistic100-photo-sphere-viewer` to make it clear that its not the original project.

mistic100:
I disagree for two reason :
1. it's against your rules of choosing a simple name
2. considering the amount of work I put in this it's more my work that his, it's just the same base idea. I would a have created a bare new repository you won't have noticed.

Beside that's what the point of making it "clear" ? People (me first) usually don't care who created what, they just want simple to use and appending the same of someone in front of the package is not simple.

jimaek:
1. What do you mean?
2. The amount of work doesnt matter. You decided to use the exact same name as an other popular project, plus you used it as the base. If you want name your project differently.
What happens when the original author decides to add his project to jsDelivr as well?

By "clear" I mean people that use the original project will look for it on jsDelivr and will expect it to be the exact same thing and not a different fork under the exact same name.

mistic100:
Considering the name on Bower does not contain "mistic100" as well as the domain name on js.org it will be as confusing.

Now as none of us wants to change it's mind I close this request. Thank you anyway.

PR #9625, opened 2016-02-07
Create jquery.loadTemplate.js

gtaasaan:
Version 1.5.0 (latest) was missing from the CDN


jimaek:
You are missing a file, please add it

PR #9635, opened 2016-02-07
Change mainfile for gridable

thasmo:
Renamed the mainfile.


thasmo:
I've renamed the mainfile which seems to be a problem regarding versioning. Isn't it possible to have different mainfiles per version? I understand that this isn't ideal - changing the name of the mainfile for new versions - but as it's still in development and not yet final, I thought this wouldn't be a problem. Can this be merged? Thanks for your feedback!

thasmo:
This is related to #9634.

jimaek:
You can have different mainfiles per version.
In info.ini leave the mainfile that is going to be added in all future versions, and in previous versions please create a new file called `mainfile` and write inside `gridable.bem.css`

jimaek:
Example https://github.com/jsdelivr/jsdelivr/commit/9559906a47d32986d8b0cda3851914f178d75642

thasmo:
@jimaek thanks for your help! :) I've added the mainfile for 0.3.0.

jimaek:
@jsdelivrbot squash and merge

PR #9657, opened 2016-02-09
Add jquery.address

leyanlo:


files/jquery.address/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
```
jimaek: There are no tags here https://github.com/asual/jquery-address/releases
The bot will fail

PR #9748, opened 2016-02-13
Add d2bk/MaterialRipple

pzmarzly:


jimaek:
No tags here https://github.com/db2k/MaterialRipple/releases
The bot will fail

PR #9802, opened 2016-02-16
Added project : similarInput.

InputSimilar:
Added project : Similar Input(similar).
version: 1.1.0
effect: Chinese search box to achieve the recommended search keywords related to accessibility.


files/similar/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "similar",
+  "repo": "InputSimilar/similar",
```
jimaek: The tags here are not numbers https://github.com/InputSimilar/similar/releases
It wont work

files/similar/update.json
```diff
@@ -0,0 +1,8 @@
...
+  "files": {
+    "include": ["similar.eva.js"]
```
jimaek: This file doesnt exist here https://github.com/InputSimilar/similar/tree/similar

InputSimilar: It has been fixed

files/similar/update.json
InputSimilar: @jimaek The author of the library has now added a release tag.

files/similar/update.json
jimaek: I didnt mean the update.json file but the `similar.eva.js` file

InputSimilar: @jimaek Are you saying `similar.eva.js` this document? I know the use of `eval()`, I use because I see 9lime are using `eval()`; but I think you are allowed to `eval()` in. Use `eval()` similar input compression can help reduce 6kb.
If the `eval()` is not allowed, please respond to me.

jimaek: I am saying that the bot uses this directory as source https://github.com/InputSimilar/similar/tree/1.1.0
And there is no `similar.eva.js` in there. So the bot will fail

InputSimilar: @jimaek Okay, I'll check.

InputSimilar: @jimaek Now it is corrected.

PR #9877, opened 2016-02-20
Add StackBlur

leyanlo:


files/stackblur/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Mario Klingemann"
+github = "https://github.com/flozz/StackBlur"
+homepage = "https://github.com/flozz/StackBlur"
+description = "Fast and almost Gaussian blur by Mario Klingemann"
+mainfile = "dist/stackblur.min.js"
```
jimaek: Just `stackblur.min.js` here

files/stackblur/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "stackblur-canvas",
+  "repo": "flozz/StackBlur",
+  "files": {
+    "include": ["dist/stackblur.js", "dist/stackblur.min.js"]
```
jimaek: Just `dist/stackblur.min.js` should be enough

PR #9900, opened 2016-02-22
Add AdminPlus Lite

lazabogdan:


jimaek:
Your project is too fresh and unstable for production use. I recommend submitting again at least when you release the first stable version.

PR #10114, opened 2016-02-25
register egjs

sculove:
I'd like to add egjs to jsdelivr CDN


files/egjs/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "egjs",
+  "repo": "naver/egjs",
+  "files": {
+    "basePath": "dist/",
+    "include": ["**/*.js"]
```
jimaek: You can remove this line, by default all files in that folder will be added

files/egjs/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "NAVER corp"
+github = "https://github.com/naver/egjs"
+homepage = "http://naver.github.io/egjs"
+description = "Set of UI interactions, effects and utilities components library."
+mainfile = "dist/eg.js"
```
jimaek: Just `eg.min.js`

files/egjs/update.json
sculove: Thank you

files/egjs/info.ini
sculove: I don't understand your intention
Our main files are eg.js and eg.min.js
What should I do?

jimaek: You should use `eg.min.js` as the parameter for mainfile

sculove:
is anyone progressing?

files/egjs/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "NAVER corp"
+github = "https://github.com/naver/egjs"
+homepage = "http://naver.github.io/egjs"
+description = "Set of UI interactions, effects and utilities components library."
+mainfile = "dist/eg.min.js"
```
jimaek: This line is still incorrect. It should be just `eg.min.js`

sculove: @jimaek OK. thank you

sculove: @jimaek I applied. please, update

PR #10152, opened 2016-02-27
Update info.ini

runestro:


jimaek:
http://www.jsdelivr.com/projects/525icons

PR #10157, opened 2016-02-27
Create info.ini

WRONGWAY4YOU:


files/cryptolist/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Dima Guzeev"
+github = "https://github.com/WRONGWAY4YOU"
```
jimaek: This needs to be the link to the project itself

jimaek:
You also need an update.json file. Follow the documentation

PR #10220, opened 2016-03-01
Inclusion of Zsocial

MaxMillion:


jimaek:
You named the folder `zsocial` but the file is named `zocial.css`
Did you make a type somewhere?

MaxMillion:
Yes. I forgot that the folder should include the ".css".  Should I redo it to reflect that?

jimaek:
No, the problem is not the .css part. Its the `s`.
One says zsocial and the other zocial. Which one is correct?

MaxMillion:
"Zocial"

jimaek:
Then fix the folder name, it says zsocial

PR #10253, opened 2016-03-03
add redux npm package

arronmabrey:
This add the "redux" npm package.

I only added the `update.json` so the bot will grab the files from npm, it wasn't super clear if I need to add the source files as well, so I didn't.

If they need to be present for the initial bootstrapping of the auto-update bot, let me know.


jimaek:
@jsdelivrbot squash and merge

PR #10306, opened 2016-03-06
Create info.ini

nemanja-stanarevic:


megawac:
The website referenced doesn't seem to ahve content and the github pages doesn't have any documentation. The project seems a bit green to host

nemanja-stanarevic:
getting there, was looking to test it out, will repost soon

PR #10554, opened 2016-03-20
Removed ngn-components in favor of chassis-components

coreybutler:


jimaek:
Sorry but we dont allow the removal of files to avoid issues for users that use them. You can just leave them and update the description if you want.

coreybutler:
I am the only user & I have the statistics to back up that claim. This will literally just clog up your environment.

coreybutler:
Thanks!

PR #10658, opened 2016-03-27
Add Quick.JS to jsdelivr

MK2018:
This is a pull request with both the info.ini & the update.json, as per the contribution guidlines, for adding Quick.JS to jsdelivr

### DIFF
--- a/files/quickjs/update.json
+++ b/files/quickjs/update.json
@@ -1,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "quickjs",
+  "repo": "mk2018/quickjs",
+  "files": {
+    "basePath": "builds/0.1a/",
+    "include": ["quick.js", "quick.min.js", "quick.css", "quick.min.css"]
+  }
+}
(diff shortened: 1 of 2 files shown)


files/quickjs/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "quickjs",
+  "repo": "mk2018/quickjs",
+  "files": {
+    "basePath": "builds/0.1a/",
```
jimaek: This format wont work, it will break the minute you release a new version.
Also there is no documentation or website.

MK2018:
I changed the file structure so that it wouldn't break, and added some basic documentation. As of right now, Quick.JS is still in development, so more documentation is forthcoming, but it'll be done very soon. Hopefully this helps.

jimaek:
I recommend waiting until the code is more stable and production ready before submitting.
Leave a comment here to review it again.

MK2018:
Thanks, I'll do a new pull request when it's ready.

PR #10695, opened 2016-03-29
Remove E2D project

jackdalton:
I'm removing this so @jtenner can use the name for [his neat project](https://github.com/jtenner/e2d).


jimaek:
We cant just remove it. It will break the websites of people using these files

@megawac Why the bot marked this as LGTM? There should have been huge warnings about removing files

jackdalton:
I'm pretty sure no websites actually use these files.

jimaek:
How about creating a new folder called `e2d.js` instead?

megawac:
Not sure, mind making a bot issue linking this thread. I'll have time to look into stuff 2 weekends from now

jackdalton:
Sounds good to me but you should check with @jtenner, he may want something
different.

megawac:
Oh @jimaek it's because the bot only cares about deletions in `<project>/<version` folders. Not about deletions in `<project>` folders

jimaek:
I see, we should probably make it care for all kinds of deletions

megawac:
@jimaek why, its not going to merge this kind of pull request and its pretty clear to the person merging there is a project being deleted

jimaek:
Someone could add 20 new files to 1 project and then remove an other project. The mod could take a look at the bot's output that there are new files and everything else is good and merge.

jtenner:
Hello Everyone!  Thank you for taking the time to help!

@megawac  is right, because the project is being deleted.  I need the e2d namespace. Nobody uses the other e2d library and it should be taken down.

I don't need anything special, and I'm also okay with using `e2d.js` as the jsdelivr namespace. It's just that it will be confusing to developers trying to use it.

jimaek:
It looks like you transferred the e2d organization to @jtenner, in this case we dont need to anything at all. We can keep all the old versions and then add the new versions. 
Plus you can update the description file.
What do you guys think?

jackdalton:
I actually deleted the organization, it should be available to @jtenner if
he desires it.

jtenner:
You guys misunderstand, @jimaek my e2d library has a different API and does different things.  Keeping the old versions of @jackdalton 's library would confuse my users.

It's a completely different project.  Sorry to be a bother.

jimaek:
@jtenner You can now submit your project like you would normally do.

jtenner:
Wonderful. Thank you!

PR #10704, opened 2016-03-30
Create outdated-browser update files

Eelviny:
Found a project over at http://outdatedbrowser.com that is publicly available, and thought it could be added to the CDN.
First time doing this, let's hope it goes smoothly.

### DIFF
--- a/files/outdated-browser/update.json
+++ b/files/outdated-browser/update.json
@@ -3,7 +3,7 @@
   "name": "outdated-browser",
   "repo": "burocratik/outdated-browser",
   "files": {
-	"basePath": "outdated-browser/",
-	"include": ["outdatedbrowser.css", "outdatedbrowser.js", "outdatedbrowser.min.css", "outdatedbrowser.min.js"]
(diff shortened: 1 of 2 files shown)


Eelviny:
Just did a bit more reading - should I remove the unminified files?

files/outdated-browser/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "outdated-browser",
+  "repo": "burocratik/outdated-browser",
+  "files": {
+	"basePath": "outdated-browser/",
```
jimaek: the path is incorrect. The correct is `outdatedbrowser`

files/outdated-browser/update.json
```diff
@@ -0,0 +1,9 @@
...
+	"include": ["outdatedbrowser.css", "outdatedbrowser.js", "outdatedbrowser.min.css", "outdatedbrowser.min.js"]
```
jimaek: You can replace all that with `*.min.*`

PR #10753, opened 2016-03-31
Patch 1

lilchow:


jimaek:
1. Its missing documentation. Where the CDN links are supposed to be used? 
2. There are no tags https://github.com/lilchow/Qualtrics-plus-plus/releases

PR #10770, opened 2016-04-01
Push New Script UTPlay

misterzik:


jimaek:
1. Only 1 project per PR is allowed.
2. UTPlay is too fresh with no documentation. I recommend waiting until its more stable before submitting

misterzik:
ahh okay, makes sense.. thanks a lot @jimaek 
I will copy readme from website. http://insanen.github.io/UTPlay

have a nice one.

PR #10807, opened 2016-04-04
Syncano update

dancio:
Hi,

What was changed:

* We are changing naming convention `syncanojs` -> `syncano` to be consistent with npm and python package
* For now we need to support both names but after few months `syncanojs` will be removed
* Added uncompressed version `syncano.js`
* Added missing `map` files for release 1.0.x 
* `update.json` && `info.ini` cleanup (new description, files, author)


jimaek:
Thank you!

PR #10810, opened 2016-04-04
Add twemoji

Jeiwan:


jimaek:
There are too many files to host them in Github. I recommend using their existing CDN by MaxCDN.

PR #10931, opened 2016-04-09
Added Flexpoint library

franciscop:


jimaek:
As I understand this PR is not needed since https://github.com/jsdelivr/jsdelivr/pull/10932 was merged

PR #10972, opened 2016-04-11
Update project jhtmlarea to 0.8.0

deltaidea:


deltaidea:
I'm not sure if I should follow those guidelines above. This PR is trying to be consistent with previous versions of the project. Can anybody comment on this please?

PR #11137, opened 2016-04-19
Add vue-i18n library

kazupon:


kazupon:
What do I need to do?

kazupon:
ping @jimaek

jimaek:
Thanks

PR #11164, opened 2016-04-20
Add timrjs

joesmith100:


files/timrjs/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "timrjs",
+  "repo": "joesmith100/timrjs",
+  "files": {
+    "basePath": "dist/",
+    "include": ["timr.js", "timr.min.js"]
```
jimaek: Just timr.min.js is enough here

joesmith100: @jimaek I want to include both files though?

jimaek: You need that in production?

joesmith100: No, but the dev version contains more descriptive errors. So would be nice to be able to offer the option to people.

PR #11174, opened 2016-04-20
add natsort

bubkoo:


files/natsort/update.json
```diff
@@ -0,0 +1,11 @@
+{
+  "packageManager": "github",
+  "name": "natsort",
+  "repo": "bubkoo/natsort",
+  "files": {
+    "include": [
+      "index.js",
```
megawac: just include `dist/` please (`"basePath": "dist/"`)

bubkoo:
I have update the config files, everything is ok now?

files/natsort/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "bubkoo"
+github = "https://github.com/bubkoo/natsort"
+homepage = "https://github.com/bubkoo/natsort"
+description = "Javascript natural sort algorithm with unicode support."
+mainfile = "./dist/natsort.min.js"
```
jimaek: Just `natsort.min.js` here

files/natsort/update.json
```diff
@@ -0,0 +1,10 @@
+{
+  "packageManager": "github",
+  "name": "natsort",
+  "repo": "bubkoo/natsort",
+  "files": {
+    "include": [
+      "./dist/"
```
jimaek: Just `dist/` here

bubkoo:
@jimaek 
@megawac 
Help me to check the files.

bradvogel:
@jimaek Natsort isn't showing up in http://www.jsdelivr.com/?query=natsort. Has this been published?

bubkoo:
@jimaek 

I have only upload `update.json` and `info.ini`, will the server auto fetch the files specified in the `update.json`?

jimaek:
There are no tags https://github.com/bubkoo/natsort/releases
thats why nothing  happened

bubkoo:
I just create one, how long does this take effect?

jimaek:
It runs every 30 minutes

bubkoo:
Thanks a lot

bradvogel:
Still not showing up - http://www.jsdelivr.com/?query=natsort. Is the tag right?

jimaek:
@MartinKolarik Any idea why its failing now? update.json looks correct

MartinKolarik:
@jimaek existing local branch would be my first guess.

jimaek:
You were right. Any idea how we can avoid this bug? I dont get why it creates a local branch but never pushes it

MartinKolarik:
@jimaek I have no idea why it sometimes fails, but I think the easiest solution would be deleting it if it exists.

jimaek:
maybe the bot can check for local branches automatically and delete them before starting the scan?
I dont think that will break anything

MartinKolarik:
yeah that's what I meant

jimaek:
ok, if its safe then lets do it

MartinKolarik:
jsdelivr/libgrabber#67

PR #11269, opened 2016-04-25
Create update.json for Codemirror

ricardopolo:
Is this the right way to include the full repo?


jimaek:
That repo doesnt host the production files but the source code. We cant use that to update

PR #11277, opened 2016-04-25
Codemirro in Bower

ricardopolo:


jimaek:
I am sorry but it doesnt match our existing structure http://www.jsdelivr.com/projects/codemirror
For example there is no min.js file

ricardopolo:
lib/codemirror.css and lib/codemirror.js arent enought?
Is abosuletly necesary to been minified?

PR #11295, opened 2016-04-26
Create info.ini

pieje:


jimaek:
you need an update.json file as well

PR #11351, opened 2016-04-28
Add monet.js

eunikolsky:


jimaek:
For production use we recommend using a minified version. Can you provide that?

eunikolsky:
hi @jimaek, I asked that question in the project: https://github.com/cwmyers/monet.js/issues/59

PR #11384, opened 2016-05-01
Create update.json

admiralsmaster:
Create json update for crypto-js to use github version (moved from google code)

### DIFF
--- a/files/crypto-js/update.json
+++ b/files/crypto-js/update.json
@@ -1,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "crypto-js",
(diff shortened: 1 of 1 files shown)


jimaek:
I cant accept this because the file structure of the repo doesnt match the current one we have live. We dont allow such big changes between minor versions. Sorry

evanvosberg:
@jimaek do you mean between minor version of crypto-js or minor version of jsdelvr?

There have been two releases of crypto-js for several year. The one on google code and the other one on GitHub. Their released structure was different all the time but the source was the same.

The google code project is no longer under development, so it makes sense to switch the dependency.

PR #11443, opened 2016-05-04
Added arg.js

fatbattk:


megawac:
I'm fine with this but the files have a version name in them (/cc @jimaek)

https://github.com/stretchr/arg.js/tree/v1.3.1/dist

jimaek:
Yes, we cant host it in this format.

PR #11469, opened 2016-05-06
eternicode bootstrap-datepicker 1.6.0 version

krzyko:


jimaek:
For the new files can you not include the unminified versions?

krzyko:
In fact almost all files are "new" because the changed location (locales subfolder) or names (bootstrap prefix in styles). Only javascript files are the same (name and placement) but they are useless without styles. Thus, maybe should I only import minified version?

krzyko:
In fact, it wold be much more comfortable to me to have both minified and unminified files hosted on cdn (uri for production and development are almost the same with only 4 chars different) but I suspect that it isn't possible because of jsdelivr policies.

PR #11513, opened 2016-05-09
[update] [owner] adding missing versions that did not participate in …

victoriafrench:
[update] [owner] adding missing versions that did not participate in autoupdate.


jimaek:
Why do you need the old versions? Only the latest should be used in production

victoriafrench:
Like Node we have a concept of LTR at Cinecove. As there are over a thousand web site installations on the previous trees, we must maintain support for them. We have been continually asked why they are not supported in the CDN, so we are adding them. The concept of "only the latest should be in production" is not true, especially with the next version that is coming. To where all the 1.x version depend on specific version ranges of modernizr and 2.x is stand-alone with no support for modernizr. Even then, 1.x will remain in LTR for a few years.

PR #11584, opened 2016-05-13
info.ini update for angular.bootstrap

isaackwan:
Proposing to change the main file to ui-bootstrap-tpls.min.js since that's the file the official bower repo points to (see: https://github.com/angular-ui/bootstrap-bower/blob/master/bower.json)


jimaek:
Sorry, but I dont want to break the websites that are already using the current mainfile.

PR #11617, opened 2016-05-14
Create info.ini

ArthurSimas1:
Created necessary documents that were missing, sorry about that. Warn me if I should add something else.


jimaek:
Sorry but this is not the kind of project we are trying to host. Its too fresh and lacks of proper documentation.

PR #11666, opened 2016-05-17
Add the Synaptic neural network library

ElGaton:
This pull request adds [Synaptic](https://github.com/cazala/synaptic), an architecture-free neural network library for node.js and the browser.


megawac:
Thanks, looks good

PR #11694, opened 2016-05-18
Add missing genetic.js reference in update.json

ElGaton:
The Genetic.js library requires __both__ files in the lib/ directory of the source repository (dist.js and genetic.js) to be redistributed; however, in my last merge request (#11668), I mistakenly specified only the first one in update.json.

This merge request adds the second one.


ElGaton:
I've noticed I need to update the info.ini file as well - please don't merge right now.

ElGaton:
I've updated info.ini and added genetic.js (after minifying it manually since upstream does not provide a reduced-size version) - this can be merged now.

megawac:
Can you revert the change to the mainfile? That may break people relying on mainfile being dist.js

ElGaton:
Done.

PR #11713, opened 2016-05-19
add files/tableheadfixer/info.ini

lai32290:


megawac:
Hi @lai32290 please submit either an update.json or a version of the library along with the info.ini

PR #11722, opened 2016-05-20
Add node-uuid

bishtawi:
Add `node-uuid`, a library that generates RFC-compliant UUIDs. Can run in web browser, its not just node.


bishtawi:
`uuid.js` is not minified. I wasnt sure how to handle that. Am I suppose to commit the minified version?

PR #11848, opened 2016-05-26
hopefully final commit

goessner:


goessner:
what is it, that 'g2.mec' does not get visible in 'jsdelivr/files' ?
must I do something else ?

PR #11862, opened 2016-05-27
Added mark.js

julmot:
Btw: I'd suggest to implement mark.js as a keyword highlighter for the [search](http://www.jsdelivr.com/?query=tes).

### DIFF
--- a/files/mark.js/update.json
+++ b/files/mark.js/update.json
@@ -1,0 +1,8 @@
+{
+  "packageManager": "bower",
+  "name": "mark.js",
+  "repo": "julmot/mark.js",
+  "files": {
+    "basePath": "dist/"
+  }
+}
--- a/files/mark.js/info.ini
+++ b/files/mark.js/info.ini
@@ -1,0 +1,5 @@
+author = "Julian Motz"
+github = "https://github.com/julmot/mark.js"
+homepage = "https://markjs.io/"
+description = "A keyword highlighter for search terms or regular expressions"
+mainfile = "mark.min.js"


julmot:
Ping @jimaek

files/mark.js/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "bower",
+  "name": "mark.js",
+  "repo": "julmot/mark.js",
+  "files": {
+    "basePath": "dist/"
```
jimaek: How about adding "include" with `*.min.*` to avoid the unminified files?

julmot: Why don't let developers choose if they want to use the minified one or not?

julmot: Ping @jimaek

jimaek: jsDelivr is focused on production usage. But if you need the unminified files then we can leave them.

julmot: Please let them for consistency as I'm offering them in all other distribution processes.

julmot:
@jimaek Thank you for merging! 

Seems like the current offered versions are quite incomplete. Does this happen due to the fact that the bot is still busy autoupdating or is there any manual help necessary?

jimaek:
We only host the latest version from each major branch. Like I said its production focused.
So the bot wont upload any older versions anymore. If you really require some older versions for some reason then you will have to add them manually

julmot:
@jimaek Is this really true? I've seen a few different projects offering all versions, e.g.
https://github.com/jsdelivr/jsdelivr/tree/master/files/markdown-it

What I'd like to have is a consistent behavior. I'm offering all versions in every distribution process, also e.g. on cdnjs: https://cdnjs.com/libraries/mark.js 
I'd like to offer jsdelivr as the preferred CDN, but can't do that if there aren't all files and all versions. 

If I understand you correctly the bot will delete previous major versions when there is a new one?

megawac:
@julmot while we will continue versions we already host for all time, we won't automatically fetch old versions

jimaek:
markdown-it has all the versions because auto-updates were enabled a long time ago and the bot simply added new releases over time. it didnt upload everything in 1 PR.

The bot will NOT delete anything. It will simply continue adding new versions.

julmot:
Alright, that said I'll have to add all previous versions manually but the bot will keep adding all new versions regardless if they are major, minor or patch?

jimaek:
Thats correct.

PR #11933, opened 2016-06-01
Add Rakam analytics

buremba:


buremba:
Is there any reason that this doesn't get any attention? Keen.IO and Snowplow already added their JS libraries so I assume that you also host JS trackers. The main repository of the project is https://github.com/rakam-io/rakam, if it's not possible to merge this PR because this is not popular enough, I can close it.

jimaek:
The website timeouts and there are no tagged versions.
I would wait until its more stable before hosting on jsDelivr

PR #11977, opened 2016-06-04
Add missing mark.js files

julmot:
As per our discussion #11862


julmot:
> Looks like the info.ini or update.json has some problems 
mark.js was renamed two times, therefore the mainfile doesn't exist before v6.0.0.

julmot:
Ping @jimaek

PR #12015, opened 2016-06-06
Updated the path

franciscop:
Updated the recommended path for Picnic CSS. Now the main file is **not** inside the `/releases`, however there is a copy there for retro-compatibility.

### DIFF
--- a/files/picnicss/update.json
+++ b/files/picnicss/update.json
@@ -3,7 +3,6 @@
 	"name": "picnicss",
 	"repo": "picnicss/picnic",
 	"files": {
-		"basePath": "releases",
-		"include": ["picnic.min.css", "plugins.min.css"]
+		"include": "picnic.min.css"
 	}
 }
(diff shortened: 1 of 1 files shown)


megawac:
Why do you want to remove plugins? Both files appear to both still be there in https://github.com/picnicss/picnic/tree/6.0.0/releases

franciscop:
I am keeping it in that folder for compatibility reasons and because apparently there's some printed books with a reference to that path (which is pretty cool). However now everything is included by default in `picnic.css`, so `plugins.css` is just a byte-per-byte copy of it.

I removed `plugins` because people were getting confused and there was too much overhead maintaining both versions. Now everything is in a single file so it's simple for both me and other developers.

franciscop:
Thanks :+1:

megawac:
Kind of sucks cause https://github.com/jsdelivr/jsdelivr/pull/12013 was merged earlier and we'd prefer to make path changes on major releases

franciscop:
I see, totally agree with semver and everything. I searched about how to do it but couldn't find it, maybe add a note to CONTRIBUTING.md?

megawac:
We'll discuss it next time there is a picnic release I suppose, @jsdelivrbot will raise a ticket about change in file structures

PR #12037, opened 2016-06-07
Updates for the 0 version

gbittmann:
The 0 version is still taking updates, but has been moved.

### DIFF
--- a/files/angular.ui-router/update.json
+++ b/files/angular.ui-router/update.json
@@ -1,7 +1,7 @@
 {
   "packageManager": "github",
   "name": "angular.ui-router",
-  "repo": "angular-ui/ui-router",
+  "repo": "angular-ui/ui-router/tree/legacy",
   "files": {
     "basePath": "release/"
   }


files/angular.ui-router/update.json
```diff
@@ -1,8 +1,8 @@
 {
   "packageManager": "github",
   "name": "angular.ui-router",
-  "repo": "angular-ui/ui-router",
+  "repo": "angular-ui/ui-router/tree/legacy",
```
jimaek: `/tree/legacy` is not part of the repo name, its a path inside the `angular-ui/ui-router` repo. So this is incorrect

gbittmann:
Thanks for helping @jimaek. Does this mean updates will need to be done manually? It's currently at 0.3.0, but the new versions aren't getting auto updated.

jimaek:
If the old versions are tagged correctly and have the same file structure then nothing is needed. The bot will keep all major versions updated. 0.x and 1.x

gbittmann:
@jimaek thanks for your continued help. I looked at the tags / releases. I'm not sure what would be wrong, but I do see several releases that have not been picked up but are months old. [https://github.com/angular-ui/ui-router/releases](https://github.com/angular-ui/ui-router/releases)
If there is something wrong on the ui-router project, I can open an issue there to see if they can fix it just let me know.

jimaek:
@MartinKolarik Any idea why the bot is not updating 0.x?

MartinKolarik:
@jimaek yeah, [we don't download](https://github.com/jsdelivr/libgrabber/blob/master/lib/project.js#L40) `0.x` releases if there already is a `1.x`. I think the reasoning behind that was that `0.x` are like beta releases and there's no point adding them if there already is a stable release.

gbittmann:
@MartinKolarik Is there any way to change that behavior in this case? Version 1 is alpha and the 0 version is stable and taking bug fixes.

MartinKolarik:
We could remove that check, but since there's a difference in file structure between 0 and 1, auto-updates will only work for one of them.

gbittmann:
I would think the 0 version would be the best fit for now since it's the stable version if you have to choose one, but I am biased. Could you go by usage metrics?

christopherthielen:
ui-router member here.

0.x releases are still the "stable" releases.  1.0 releases should be finalized in a few months.

github.com/angular-ui/ui-router is no longer the right place to be pulling releases from.  Can jsdelivr use npm packages, or github only?

If github only, it should be fetched from our bower releases at http://github.com/angular-ui/angular-ui-router-bower which should have the same package structure as before.


---

We have changed our distribution to allow angular 1 and angular 2 code to co-exist in a single github repository.   From the single monolithic repository, we're publishing 3 different packages:

- ui-router-core
- ui-router-ng2 (angular 2)
- angular-ui-router (angular 1)

We tag the overall repository when releasing, but we no longer publish directly from the tag.  We now build npm packages and bower packages and publish separately from the repository.

PR #12060, opened 2016-06-08
Added Latest LiveFyre JS Library

itseasy21:
From Community to Enterprise, the **Livefyre Apps Plugin** infuses your website with real-time social content to increase engagement and drive traffic.


files/livefyre/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "LiveFyre"
+github = "https://github.com/itseasy21/jsdelivr"
```
jimaek: This needs to be the project's official repo and the fork of jsdelivr

itseasy21: But the project is not on github, then how can it be added ?

jimaek: Where is it hosted then?

itseasy21: I got it from : http://zor.livefyre.com/wjs/v3.0/javascripts/livefyre.js
It is hosted on livefyre servers, which is the comment system that i am using, but they dont have a cdn and neither their JS is available on any public CDN.

jimaek: It doesnt look like its an open source project. We dont just host js files. It needs to have a project page and a license.

PR #12281, opened 2016-06-20
Update project babylon.js

vousk:
Source files are : 
https://github.com/BabylonJS/Babylon.js/blob/master/dist/babylon.2.4.js
https://github.com/BabylonJS/Babylon.js/blob/master/dist/previous%20releases/babylon.2.3.js


vousk:
This is not an inconsistency between versions. 
The babylon.min.js file is the correct minified complete version, the babylon.core.min.js file is only the root part of the lib. Usually I only upload the complete version but for 2.3 I was asked to explicitly put also the core file. Since this file may be used sometimes, I can't remove it. 
The real 2.4 file to upload is babylon.min.js, not babylon.core.min.js.
How can I bypass the bot checking ?

jimaek:
Cant check right now, traveling. @MartinKolarik Can you please check?

MartinKolarik:
@vousk it seems this PR also adds `2.3/babylon.min.js` which already is on jsDelivr?

vousk:
@MartinKolarik Yup, my bad. My previous PR added this 2.3 file to the jsDelivr main repo but I forgot to update my own fork on github, so in my forked repo the file wasn't there, that's why yesterday I also added 2.3 final with 2.4 final (without previously checking the jsDelivr original repo...).
I will close this PR and request a new one with 2.4 file only.

PR #12283, opened 2016-06-20
Add map-icons.js

ahirmayur:
Map Icons is an icon font for use with Google Maps API and Google Places API using SVG markers and icon labels.

Map Icons makes Google Maps Markers dynamic with control over shape, color, size, and icon easily changed using options in the marker object as well as simple SVG Notation and CSS.


jimaek:
@MartinKolarik @megawac Can one of you please check this and merge if its good? Im traveling and cant review it (internet super sucks)

PR #12331, opened 2016-06-22
Update project babylon.js to 2.4 final

vousk:
Source file is : https://github.com/BabylonJS/Babylon.js/blob/master/dist/babylon.2.4.js


vousk:
@MartinKolarik @jimaek Ok, I closed the previous PR, removed the 2.3 useless commit and create this new PR with only 2.4 file. There still is a falsly failed bot ckeck because the babylon.core.min.js file was only sent for 2.3 but (probably) won't be needed for 2.4

PR #12387, opened 2016-06-24
Add requarks repo

NGPixel:


megawac:
I'm going to defer this to @jimaek but we generally host libraries and not entire applications.

NGPixel:
I've limited it to the **/assets** folder only, which contains the client js and css files for the application, not the whole application itself.

To put it simply, it's like hosting hosting the client files only of Bugzilla. So any organization installing Bugzilla on their server would benefits from having users fetching the client files from the CDN.

Thanks,
Nick

PR #12460, opened 2016-06-29
Push bootstrap.star rating

DanCollinz:


megawac:
Close, just need to fix that quote mark now to be "

DanCollinz:
@megawac changed the fancy quotes, thanks for looking!

DanCollinz:
@megawac I think this is ready for merge, but let me know if there is anything else I should do to move this along

PR #12490, opened 2016-06-30
Adding animate css

DanCollinz:


files/animate-css/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "animate.css",
+  "repo": "daneden/animate.css",
+  "files": {
+    "basePath": "/"
```
megawac: can you add `include: ['animate.min.css', 'animate.css']`

PR #12492, opened 2016-06-30
Adding twig

DanCollinz:


megawac:
I don't think this library will work for auto updating as they don't upload the compiled version of twig.

If you want to compile it with webpack feel free to submit the built file(s)

DanCollinz:
ThanksI was't sure if this would work, I'll compile the file and see how that goes. Thanks for the speedy feedback!

megawac:
The compiled version isn't in recent versions of twig. I'd suggest including only the info.ini and compiled js file

PR #12494, opened 2016-06-30
jquery.payment - Add all files from project to include minified version as well

delmicio:
Get all files from project not just the main unminified version.


jimaek:
why?

delmicio:
@jimaek to include the minified version. Right now only the unminified version is grabbed by the boot.

delmicio:
@jimaek why is not the minified version available yet in http://www.jsdelivr.com/projects/jquery.payment? is the bot not getting this change?

jimaek:
because there are no new versions released. The change will apply only to new versions and not to old.

delmicio:
A new version may not come soon, months even.
Can it be applied to the latest at least?

jimaek:
It must be done manually, if you want you can open a PR adding that file

PR #12497, opened 2016-06-30
Update Materialize with 0.97.x versions

Dogfalo:


Dogfalo:
Hi @jimaek,
I want to add some older versions for Materialize, but the bot complains that the project does not have an `info.ini` file which it does.

jimaek:
Why do you need the old versions?

Dogfalo:
Currently the project has a `info.ini` which someone else submitted, but has no actual versions yet. People have been asking for me to put it on Jsdelivr so I wanted to backfill some versions.

PR #12529, opened 2016-07-03
Update eruda update.json

surunzi:


files/eruda/update.json
```diff
@@ -3,7 +3,6 @@
   "name": "eruda",
   "repo": "liriliri/eruda",
   "files": {
-     "basePath": "dist/",
```
jimaek: Why? The current version has the dist/ folder
https://github.com/liriliri/eruda/tree/v1.0.0/dist

surunzi: I just removed the dist/ folder in the latest version published on npm.

PR #12559, opened 2016-07-05
adding library

IPENGR:
first time adding the lib to jsDelivr , this project is used for educational purpose,
We are releasing 0.2 version of our project we need some cdn for our js files


IPENGR:
how can I utilize my library

IPENGR:
can I, know when the lib will approved?

megawac:
Hi @IPENGR jsdelivr is generally used for hosting open source code that is usable in multiple applications. This looks like a custom application for a specific site.

We also require libraries have some documentation and maturity before we accept them. With these points noted I'll have to reject this request

Further we would need you to not use `0.2` as the basePath but something like `release/` or `dist/`.

PR #12574, opened 2016-07-06
Patched Noto Sans KR Webfonts & Main file path in info.ini

korusdipl:
1. Due to the rendering error (baseline floating) on some browsers, it gave different experience/errors across browsers. Patched.

2. The main css file at the dist folder doesn't show up properly previously, changed info.ini pathinfo to resolve the issue.

* Due to the nature of this resource, I'd rather replace the files than incrementing versions and copy contents 'as always'

Thanks for your attention!


megawac:
can you create a new version instead of updating the old one? how about changing version to `1.1/` or `2/`

korusdipl:
I've done it in the above way because of the following scenario.

I introduced users to use the 'latest' url, so by this way users might experience difficulty when using other fonts than Noto Sans if I update by changing the number for those files.

For example, if someone used 
/1/corps/daum/Daum/Daum-Regular.woff - (a)
as latest and I update Noto Sans font in /2/ folder (or so), one might lose access (link broken) through the example full URL when accessing through 'latest' (tell me if I am wrong about this)

Elaborated, if I update the URL by
/2/corps/google/NotoSansKR/NotoSansKR-Black-Latin.woff
then one can no longer access Daum-Regular.woff by /latest/corps/daum/Daum/Daum-Regular.woff

Or, are you suggesting changing the whole file from 1 to 2 (sort of?) I believe that is too wasteful.

Thanks for your attention again :)

megawac:
Hey, yeah I mean you should leave version 1 as is and create a new version >1. Anyone using latest URLs will use the new version, people using /1 URLs will experience no changes

korusdipl:
Thanks for the quick reply! but,
What about people using /latest URL but expecting to get from /1 URL?
My concern came from that.

megawac:
The `/latest` url is inherently unsafe, you are opting into possible major changes.

This pr also would break anyone relying on the previous 1 url as the version would change under their feet

korusdipl:
Ah, all right, I finally understood your point. 
I will change the PR accordingly and post again when I get back.
Thanks for all replies!

PR #12641, opened 2016-07-09
Update update.json for gmaps.js

yumeixxx:


jimaek:
Why do you need this file?

PR #12650, opened 2016-07-09
Add angular.ng-xhr-promisify

tiago:


jimaek:
Your project has no documentation and only 1 commit. I would wait until it becomes more stable before submitting.

tiago:
I specifically looked into the readme for restrictions on library publication, but found none. In the meanwhile, usage documentation is already there.
Anyway, feel free to close this PR if it doesn't fit your requirements. I'll then re-submit the lib at a later time when it's mature enough.

PR #12678, opened 2016-07-11
Add hello.js proyect

delmicio:
Hopefully the files/basePath is correct...

### DIFF
--- a/files/hello.js/update.json
+++ b/files/hello.js/update.json
@@ -3,6 +3,7 @@
   "name": "hello.js",
   "repo": "MrSwitch/hello.js",
   "files": {
-    "basePath": "/dist/"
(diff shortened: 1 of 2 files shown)


files/hello.js/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "hello.js",
+  "repo": "MrSwitch/hello.js",
+  "files": {
+    "basePath": "/dist/"
```
jimaek: Just `dist/`

jimaek: Also add an `include` parameter with `*.min.js`

delmicio: done.

PR #12680, opened 2016-07-12
Breinify API Library

breinify-stage:
First public version of the Breinify API Library (JavaScript)


pmeisen:
Hi,

just wondering if there are any problems, or when we can expect the library to be available at jsdelivr? Thanks!

megawac:
@pmeisen you just need to start adding `git tags` to the brein-api repo for this to be accepted

pmeisen:
Thanks for the info, did that a couple of hours ok. When does the jsdelivr system scan the repository and adds it to the index?

Thanks a lot for the quick replies, appreciate that a lot! Great service!

PR #12704, opened 2016-07-13
Add webfont spoqa han sans

the1900:
This font is better than NotoSans and I think people will like it.

### DIFF
--- a/files/font-spoqa-han-sans/info.ini
+++ b/files/font-spoqa-han-sans/info.ini
@@ -1,0 +1,5 @@
+author = "Spoqa (spoqa.com)"
+github = "https://github.com/the1900/spoqa-han-sans"
+homepage = "http://spoqa.github.io/spoqa-han-sans/"
+description = "spoqa han sans is a sans-serif type web font for korean, english, japanese"
+mainfile = "SpoqaHanSans.css"
--- a/files/font-spoqa-han-sans/info.ini
+++ b/files/font-spoqa-han-sans/info.ini
@@ -1,5 +1,5 @@
 author = "Spoqa (spoqa.com)"
-github = "https://github.com/the1900/spoqa-han-sans"
+github = "https://github.com/spoqa/spoqa-han-sans"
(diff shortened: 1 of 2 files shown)


the1900:
Why jsdelivr hasn't copied my spoqa-han-sans [project](https://github.com/the1900/spoqa-han-sans) to CDN?? wasn't it finish what I need to do ?

files/font-spoqa-han-sans/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Spoqa (spoqa.com)"
+github = "https://github.com/the1900/spoqa-han-sans"
+homepage = "http://spoqa.github.io/spoqa-han-sans/"
```
megawac: wait, why are you referencing a different website?

megawac:
@the1900 why don't you submit these changes to the original spoqa repo instead of forking and taking spoqa's assets?

files/font-spoqa-han-sans/info.ini
the1900: Well I haven't thought about it.

Anyway, As you commented. I requested spoqa to add it in JSDelivr

the1900:
@megawac  If you don't mind can you answer me?
I added info.ini and update.json. After this I've requested merge so are these all? even I don't use bower. I forked jsdelivr and add files on patch-1 branch so I think JSdelivr should copy my spoqa-han-sans repo. 

If you canceled merge becase it is not the original repo then please notify me. I want to know am I right to add files to JSDelivr CDN. I thought it is all automatic when using github wich contribute.md recommended Now I am confused.

megawac:
@the1900 you can submit a project on behalf of another repo, its just odd to submit a fork of another project under the same name. I don't think we have precedent here

the1900:
@megawac Oh this is what you said, right??

yeah I misunderstood the process. I thought only my repo could be submitted to JSDelivr.

megawac:
Thanks looks good

PR #12797, opened 2016-07-19
add new package

Syncfusion-JavaScript:


Syncfusion-JavaScript:
Hi Team,

1. We have committed the info.ini file in this repository. But the jsDelivrbot is reported like "syncfusion-javascript has no info.ini file buddy" . Find the committed file path for your reference. 

https://github.com/Syncfusion-JavaScript/jsdelivr/blob/master/files/syncfusion-javascript/info.ini

2. It seems to be report some concerns about to commit .GIF, .less and .cur files. but we need to commit this in jsDelivr repository.

megawac:
Yeah, the bot gets confused when there are a lot of files

PR #12823, opened 2016-07-20
Add bytes.js

hairyhenderson:


megawac:
I don't think this module will work as it relies on CJS exports for node and is not a browser file

PR #12842, opened 2016-07-21
Patch 2

theducksnuts:


megawac:
This looks like custom code for a single site. We try to only host code which is usable for more than one use case.

theducksnuts:
Hi Graeme.

 

This code is for my Shopify App – there is 500 shops using it, and in the last month there has been 500,000 requests for this file. Please advise if it does not suit still and I will make other arrangements.

 

Kind Regards,

Dave Martin
_______________________________________________________________

RapidSites | Designer & Developer

* Mob +61 (0) 455 666 414
 <mailto:david.martin@rapidsites.com.au> david.martin@rapidsites.com.au  |  <https://www.rapidsites.com.au/> www.rapidsites.com.au

 This email and any files transmitted with it are confidential and intended solely for the use of the individual or entity to whom they are addressed. This message contains confidential information and is intended only for the individual named. If you are not the named addressee you should not disseminate, distribute or copy this e-mail. If you are not the intended recipient you are notified that disclosing, copying, distributing or taking any action in reliance on the contents of this information is strictly prohibited.

megawac:
We typically only host libraries (i.e. stuff that can be used across multiple projects with some documentation. This doesn't seem to be the case with ShopifyApp

theducksnuts:
Ok mate. I guess I’ll look elsewhere.

PR #13367, opened 2016-08-22
Add intense-image library

brandomeniconi:


megawac:
sorry for delay, you'll need a git tag before this will show up on jsdelivr

PR #13402, opened 2016-08-23
Create update.json

delmicio:
I'm not sure if this is the correct way for this project. 
May be could be grabbed from https://github.com/cdnjs/cdnjs/tree/master/ajax/libs/chosen

### DIFF
--- a/files/chosen/update.json
+++ b/files/chosen/update.json
@@ -1,0 +1,5 @@
+{
+  "packageManager": "npm",
+  "name": "chosen",
+  "repo": "chosen-js"
+}
--- a/files/chosen/update.json
+++ b/files/chosen/update.json
@@ -2,5 +2,8 @@
   "packageManager": "npm",
   "name": "chosen",
   "repo": "chosen-js",
-  "files": "{}"
+  "files": {
+    "include": ["*"],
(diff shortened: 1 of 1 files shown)


delmicio:
@jimaek could you help me a bit here? please

megawac:
You need to specify `files` in the update.json

delmicio:
@megawac  `files` is optional.
I've never worked with npm, I don't understand how to find the files there.
I wonder how can cdnjs bot have it updated there.

@pfiller, @kenearley, @stof, @koenpunt, @tjschuck.

megawac:
files is not optional. `basePath` and `include`/exclude are optional. Anyway your issue now is that files is a string and not an object.

As for the files to include do an npm install for chosen and list the files you want included on the cdn

delmicio:
@megawac  Thanks you! I've done that and found the files. Although no minified version is provided this way :-1:

files/chosen/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "npm",
+  "name": "chosen",
```
megawac: I think you want this name to be chosen-js, `chosen` on npm looks like the wrong thing

PR #13424, opened 2016-08-24
Chart.min.js v2.2.1

lewisgoddard:
The autoupdater doesn't work because it moved and it doesn't include the dist files anymore.


lewisgoddard:
`Chart.bundle.min.js` was introduced since the auto-update link was broken, and is a recommended distribution file by them.

PR #13439, opened 2016-08-25
Delete 1.0.0/geniuslink.min.js

modethirteen:
/mindtouch-geniuslink/1.0.0/geniuslink.min.js was incorrectly built on our side, and it being a 1.0.0 release it would be in poor form to release it incomplete. We have confirmed that no one is using this release.


megawac:
How did you confirm this wasn't being used?

modethirteen:
@megawac this library, while open source, only serves @MindTouch customers. We have yet to announce it to them as officially supported (we are waiting for the correct 1.0.0 release to do so). Using this library at the moment outside of any sort of beta development isn't supported under any SLA.  The only customers that are part of the beta program have confirmed they aren't using this version of the library.

PR #13490, opened 2016-08-28
Patch 1

SimeoneVilardo:
Added the Bootstrap plugin "bootstrap-toggle" with the standard file structure.
Here more info: http://www.bootstraptoggle.com/


megawac:
I think you forgot to set the basepath to `js/` https://github.com/minhur/bootstrap-toggle/tree/2.2.2

PR #13505, opened 2016-08-29
rename dracula to match npm package name

strathausen:
Apparently, the auto updater failed due to the name mismatch with npm. This commit should fix that!


megawac:
Are you sure that's the issue? https://github.com/strathausen/jsdelivr/blob/f7b7d5b627246c5b1e706e1d4ea7f8ba1ca401dc/files/graphdracula/update.json#L2 says we're using github not npm

strathausen:
Actually, I'm not sure, but https://github.com/jsdelivr/jsdelivr/pull/13466 says:

> Looks like the info.ini or update.json has some problems
> - dracula has no info.ini file buddy

Any idea what else that could be? The info.ini has content: https://github.com/jsdelivr/jsdelivr/blob/master/files/dracula/info.ini

megawac:
oh, that is an api bug. We're working on that one

strathausen:
Ah cool 👍 This means I should close this PR?

megawac:
yea, lets close it. Moving discussion to https://github.com/jsdelivr/api/issues/130

PR #13613, opened 2016-09-05
Add mdbootstrap

mattonit:


megawac:
For the auto updating to work you'll have to start using `git tags` or `github releases`

PR #13747, opened 2016-09-12
Add OpenLayers Interactive Overview

jonataswalker:
Thanks for your work.

### DIFF
--- a/files/openlayers.interactive-overview/update.json
+++ b/files/openlayers.interactive-overview/update.json
@@ -1,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "ol3-interactive-overview",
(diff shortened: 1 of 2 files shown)


files/openlayers.interactive-overview/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
```
megawac: need to start tagging releases with git before this will work

jonataswalker: Oh I see, now there's [a released one](https://github.com/jonataswalker/ol3-interactive-overview/releases/tag/v0.1.0).

PR #13830, opened 2016-09-16
paste-image

redgeoff:


files/paste-image/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "bower",
+  "name": "paste-image",
+  "repo": "redgeoff/paste-image",
+  "files": {
+    "basePath": "dist/",
```
megawac: is dist present on bower only?

redgeoff:
Yes, that is correct. If you do a `bower install paste-image` you'll see
the dist directory and the built js files. Better to keep build files out
of github when possible ;)

megawac:
roger

PR #14020, opened 2016-09-27
Updates to 3.5.41

etsms:
* adds new visuals for the secure icon
* adds new "connection" timeout for transvault devices


etsms:
^^ Ignore.

PR #14026, opened 2016-09-28
Added gateway lib for the original web2web concept (https://elendirx.github.io/web2web/)

elendirx:


megawac:
you'll need to create a git tag or release before auto updating will work

PR #14035, opened 2016-09-28
Create info.ini

yieme:


yieme:
add feathersjs

yieme:
should be feathersjs/feathers-client/tree/master/dist

PR #14058, opened 2016-09-30
Headstart-CSS

RoteMinecraftIL:


megawac:
We already host entypo. Perhaps rotem only?
http://www.jsdelivr.com/?query=entypo

RoteMinecraftIL:
OK. Removing entypo.

PR #14243, opened 2016-10-06
Update project oppia to 0.0.2

seanlip:
Adding version 0.0.2 of oppia.

Source of the file: https://github.com/oppia/oppia/blob/develop/assets/scripts/oppia-player-0.0.2.min.js


files/oppia/info.ini
```diff
@@ -1,5 +1,5 @@
 author = "Sean Lip"
 github = ""
```
megawac: want to set this field as well?

seanlip: Thanks, good catch! I made the modification.

PR #14257, opened 2016-10-07
Create info.ini

sharad-iise:


megawac:
wrong folder. You need to include either files or an update.json along with the info.ini

PR #14272, opened 2016-10-07
Patch 1

fernandosouza:


megawac:
You'll need to create a git tag or github release before auto updating will work

PR #14350, opened 2016-10-13
Added complete 3.3.1 modernizr minimized and not files

marioja:
This includes a file with 258 browser feature check capability.


marioja:
I think it is ready now.

PR #14412, opened 2016-10-18
Update project crypto-js to last github version

albuic:
Update the path of the project as the google project is dead. Now using github fork of @brix which is the most used one.


albuic:
Just update the author for the bot

PR #14417, opened 2016-10-18
Create info.ini for lazzy.js

eldelentes:


megawac:
needs a git tag or release before auto update will work

eldelentes:
@megawac Thank you, Yeates, I added the release

PR #14443, opened 2016-10-20
Add Library Spotter_Tools

exactsales:


megawac:
You're library is going to need some documentation and a tagged release before we can proceed.

exactsales:
Hi, it's done.

This library is just some external tools that we would like to add to our site using your CDN. That's why we don't have a good documentation.

megawac:
jsdelivr is an open source CDN. Libraries hosted here should be open source, documented and applicable to more than just a single site. Sorry.

exactsales:
Hi,

I got it. Sorry about that.

PR #14447, opened 2016-10-21
Update project graphql-toolbox to 1.0.1

tlvenn:


tlvenn:
Hey @megawac, sorry to bug you again, had to push a fix... :(

I promise, next update will include an update.json so we can automate this.

Thanks a lot in advance and have a great WE.

PR #14448, opened 2016-10-21
Update es6-promise/update.json

limonte:
Include all important JS files (https://github.com/stefanpenner/es6-promise#downloads)


megawac:
thanks

PR #14467, opened 2016-10-26
Create info.ini

xXP4trickXx:


megawac:
you will also have to upload an update.json or some files in addition to the info.ini

PR #14471, opened 2016-10-28
Update update.json

estebanrfp:


estebanrfp:
I have seen that the name of the project was not the same as the submited folder and I've fixed it

PR #14481, opened 2016-10-31
Add project blocksovetnik

nofuel:


megawac:
you'll need to create a git tag on your repo before auto updating will work. Otherwise looks good

Issue #13217, opened 2016-08-13
Too many libraries not updated


ddddgh:
It used to be a good source of libs. At least for the most commonly used jquery. Slowly when i started using more "uncommon" files it starts to show it's weakness.

It's really not that uncommon. Just a few example of what i'm using

Pace.js lots of version missing

Zepto.js latest is 1.2.0

Normalize.css latest is 4.2.0

Prefixfree

Etc...

jimaek:
The new backend is live. It pulls files directly from npm and github https://github.com/jsdelivr/jsdelivr#usage

PR #14655, opened 2016-11-05
Update info and add update.json to normalize.css

lenovouser:
#13217

(It seems like there is no minified version available anymore 😞 )


lenovouser:
In case anyone runs into the ASCII problem too, here is how you find them in your file:

```
cd ~/Development/jsdelivr/
grep -P -n "[^\x00-\x7f]" files/<file.ext>`
```

megawac:
Cheers thanks @lenovouser

PR #14670, opened 2016-11-05
Patch 1

Petrovich-Z:
Add my script to your cdn, please.


megawac:
fyi i merged, but you'll need to create a git tag before auto updating will take place

Petrovich-Z:
What I'm doing wrong? I added a tag to my project yesterday, but my project folder contains only info.ini and update.json.

megawac:
gss4tilda.min.js doesn't exist in the root directory. You should add basePath `src/js`

Petrovich-Z:
Thank you.

PR #14679, opened 2016-11-06
Add selectshare assets to jsdelivr

guuibayer:


megawac:
if this is a chrome extension is there any reason to host on a cdn. It's generally better to bundle assets

PR #14825, opened 2016-11-17
Add vue2-filters

freearhey:


megawac:
you'll need to publish a git tag before this will work fyi

freearhey:
thanks for the help!

PR #14828, opened 2016-11-17
Update update.json

Dog2puppy:


megawac:
@Dog2puppy the actual issue is that the file names are capitalized

Dog2puppy:
Thanks @megawac. It's those small mistakes that are hard to find.

PR #14865, opened 2016-11-19
WIP | Fix is.js not being updated

lenovouser:
I don't really see why it is not being updated, but there are lot's of releases after `v0.52`, which is the only version available on the CDN at the moment.

@megawac do you have any idea? Maybe libgrabber is just stuck? Because I can't really see why it isn't pulling the latest changes.

[arasatasaygin/is.js/releases](https://github.com/arasatasaygin/is.js/releases)

### DIFF
--- a/files/is.js/info.ini
+++ b/files/is.js/info.ini
@@ -1,5 +1,5 @@
 author = "arasatasaygin"
 github = "https://github.com/arasatasaygin/is.js"
-homepage = "http://arasatasaygin.github.io/is.js/"
+homepage = "http://is.js.org"
 description = "Check types, regexps, presence, time and more..."
 mainfile = "is.min.js"
--- a/files/is.js/update.json
+++ b/files/is.js/update.json
@@ -3,6 +3,6 @@
   "name": "is.js",
   "repo": "arasatasaygin/is.js",
   "files": {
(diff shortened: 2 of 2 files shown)


megawac:
The reason is is because the latest tag jsdelivr recognizes is `0.52`. This is treated as `0.52.0` which is greater than `0.9.0` (is.js latest release) causing libgrabber not to pull it. I think this will require a manual upload of 0.9

http://www.jsdelivr.com/projects/is.js

lenovouser:
What about deleting `v0.52` and uploading the first version that is available? ( Which would be `v0.2.0` **=>** [arasatasaygin/is.js](https://github.com/arasatasaygin/is.js/releases?after=v0.2.1) ) I think it should automatically add the other versions after that. I suppose that `v0.52` actually was uploaded by mistake because it doesn't exist in the original repository.

megawac:
i assume so too, but that could break urls for sites relying on it. I think we have to respect that its there and manually upload new versions until 1.0.0 comes out

lenovouser:
Ah, true that. Are rewrites possible? Because we could do what I just said above and then add a `301` from `0.52` to `0.5.2` which seems to be the actual version.

PR #14875, opened 2016-11-20
Jquery Mosaic (Minified)

manfuku:
Manually added minified files as these aren't available in the original repository (https://github.com/tin-cat/jquery-mosaic)


megawac:
You need to place them under a version. E.g. `files/jquery-mosaic/v1/<file>`

PR #14888, opened 2016-11-21
Add ndgevnt

janplexust:


megawac:
I'm not sure what this project is but projects hosted on jsdelivr are supposed to be open source libraries which are reusable between different projects. This means they should have at least some documentation (a readme/wiki page).

Feel free to reopen when ndgevnt has matured a bit

PR #14958, opened 2016-11-25
[request] add check-browsers

perfectacle:


megawac:
you'll need to create a git tag before the initial version will be picked up

perfectacle:
hey, what is git tag...?
how can I add it...?

PR #15126, opened 2016-11-30
Patch 1

krzysztofmat:


files/parafiadywity/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "parafiadywity",
+  "repo": "krzysztofmat/parafiadywity",
```
megawac: I can't find https://github.com/krzysztofmat/parafiadywity

krzysztofmat: Thank you for this comment. On the way to fix this.

megawac:
Sorry, I didn't notice when this was first requested, but this looks like custom code for a specific website. We only host open source libraries on jsDelivr

PR #15277, opened 2016-12-10
Adding AngularJS 1.6.0

ashlinallen:


ashlinallen:
@megawac There appears to be an actual difference in the files located in i18n/. You can confirm for yourself in the zipped libs at https://code.angularjs.org/1.5.8/ and https://code.angularjs.org/1.6.0/.

PR #15287, opened 2016-12-11
Dummy.js

labsvisual:
Added files to create a CDN fork for Dummy.js


megawac:
don't mind these messages fyi, testing some code ^^

PR #15349, opened 2016-12-14
Add Gravatar.js

CosmicWebServices:
Official Page: https://github.com/CosmicWebServices/gravatar.js


CosmicWebServices:
Wow that was harder than it thought it would be *wipes sweat off of forehead*

Issue #15568, opened 2016-12-30
d3.js project not picking up new versions


bradvogel:
Hi, this is a question that I asked in Gitter but didn't get a response. So I'm asking here hoping someone can help.

I'm confused why libgrabber stopped picking up D3.js updates at version v3.5.17, when the latest is v4.4.0 in D3's github repo: (https://github.com/jsdelivr/jsdelivr/tree/master/files/d3js). It has an auto-update file (https://github.com/jsdelivr/jsdelivr/blob/master/files/d3js/update.json) and the repo is properly tagged in git.

MartinKolarik:
It's just a guess but try to send a PR to change `repo` in update.json from `mbostock/d3` to `d3/d3`

bradvogel:
Thanks! Filed https://github.com/jsdelivr/jsdelivr/pull/15569. Will close this.

PR #15569, opened 2016-12-30
Point D3js at the npm distribution

bradvogel:
Fixes https://github.com/jsdelivr/jsdelivr/issues/15568. The D3 repo was moved and broken up into several different repos, so build files no longer exist in the repo, but they are built when publishing to npm. Therefore it's easier to just point JSDelivr at the npm distribution where the build files exist.


bradvogel:
ping @megawac :)

PR #15662, opened 2017-01-06
Add inferno

jdalton:
This PR adds [inferno](https://infernojs.org/), the extremely fast React-like JavaScript library for building modern user interfaces.

### DIFF
--- a/files/inferno/info.ini
+++ b/files/inferno/info.ini
@@ -1,0 +1,5 @@
+author = "Dominic Gannaway"
+github = "https://github.com/infernojs/inferno"
+homepage = "https://infernojs.org/"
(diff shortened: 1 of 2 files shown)


jdalton:
\cc @trueadm to sanity check.

megawac:
gracias

jdalton:
Ah, I think there's a bit of a bug:

It says the main file is this, which points to a 404
https://cdn.jsdelivr.net/inferno/1.1.1/packages/inferno/dist/inferno.min.js

jdalton:
The "mainfile" entry needs to be changed to the based-path of:
```txt
mainfile = "inferno.min.js"
```

megawac:
Ah thanks, will fix

PR #15809, opened 2017-01-16
Update KaTeX to 0.7.0

as-com:


as-com:
we really need to get through these pull requests...

megawac:
what do you mean?

megawac:
I'm pretty sure 90% of the open ones are never going to be merged. About 80 of them are bot pull requests that don't seem safe to merge, then there are like 60 or so user requests that are stagnant

PR #15810, opened 2017-01-17
Add w3schools css

herohamp:


as-com:
Hampton, make sure you follow the instructions to properly add a new project: https://github.com/jsdelivr/jsdelivr/blob/master/CONTRIBUTING.md#advanced---manual

You need to have a info.ini file and a separate folder for each version. You'll also want to add a properly minified version.

PR #15823, opened 2017-01-17
Patch 1

klhurley:


files/jquery.simplemodal/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Kenneth Hurley"
+github = "https://github.com/klhurley/jsdelivr"
```
megawac: this repo won't work. It should be a repo containing the source code of jquery-simple-modal.

PR #16051, opened 2017-01-29
Info.ini

Alecgcast:
CDN for animedescarga


megawac:
I'm sorry, but this request looks like its requesting a library that is specific to a single application and not an open source library. The purpose of jsDelivr is to host open source libraries which can be reused across multiple projects. As such, we generally require new projects to have at minimal: some documentation, a source repository or website to download the code and some general reusability.

PR #16167, opened 2017-02-05
Add: moesif-browser-js lib

dgilling:
located at https://github.com/Moesif/moesif-browser-js


dgilling:
Added tag for current version v1.0.0. Any other action needed?

PR #16283, opened 2017-02-10
Kevguy patch 1

kevguy:


kevguy:
tagrhead is added but files on https://cdn.jsdelivr.net are not present(try https://cdn.jsdelivr.net/tagrhead/1.0.3/Tagrhead.min.js)

megawac:
The issue is on the repo the file names are Tagrhead and in this PR they are TagrHead

https://github.com/kevguy/TagrHead/tree/v1.0.4/lib

PR #16288, opened 2017-02-10
Project has been recreated

xavierfaucon:


megawac:
why are you deleting existing files. This could break any websites depending on those files.

xavierfaucon:
You are right ; I know it's not good practice but since repo has been started again from scratch I thought it'd be better to remove everything.

What would you advise? keep those old files and bump to a higher version in the current repo?

megawac:
Yes for sure. Start with a new major version would be my advice

PR #16298, opened 2017-02-10
Create info.ini

stormyqq:


megawac:
an initial version or an update.json must be included as well.

PR #16299, opened 2017-02-10
Create bootstrap.min.js

stormyqq:


stormyqq:
but I just submitted a info.ini file, so how come you said it doesn't have a info.ino file?

megawac:
The info.ini and initial version should be submitted in the same pull request

PR #16322, opened 2017-02-13
Add hola_player

sreym:


files/hola_player/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "npm",
+  "name": "@hola.org%2fhola_player",
```
megawac: ` npm info @hola.org%2fhola_player` gives me an error. Can you double check this

sreym: I tested it via libgrabber and it was only way to use it. How I can pass npm-name like this:
'@hola.org/hola_player'?

megawac: ah okay, lets give it a shot I'm not sure

PR #16340, opened 2017-02-13
Patch 3

pelletierjy:
Hey!  Thanks for helping!  
We need this angulartics.

### DIFF
--- a/files/angulartics/update.json
+++ b/files/angulartics/update.json
@@ -1,7 +1,7 @@
 {
   "packageManager": "github",
   "name": "angulartics",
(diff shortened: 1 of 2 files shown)


files/angulartics/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "angulartics",
+  "repo": "angulartics/angulartics",
+  "files": {
+    "include": ["angulartics.js", "angular-moment.min.js"]
```
megawac: I don't see those files https://github.com/angulartics/angulartics/tree/1.4.0

PR #16491, opened 2017-02-22
syncano-client-js 

tomsonpl:


files/syncano-client-js/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Eyedea"
+github = "https://github.com/syncano/syncano-client-js"
+homepage = "http://syncano.io/"
+description = "Interact with your Syncano Sockets."
+mainfile = ""lib/index.js","
```
megawac: whoops wasn't paying attention. This needs to be fixed

tomsonpl: @megawac  should be`     "lib/index.js"   ` ?

megawac: yeah, that looks right

tomsonpl: it was like this and I got a response to fix it. Check your autoresponder ;p

tomsonpl: thanks, I created another PR for that, couldn't find out how to do this here.

megawac: ah, it got confused because of the trailing comma https://github.com/jsdelivr/jsdelivr/pull/16491/commits/a6aa33f6b6d29f01147443aee2a863eb0f747986#diff-4e1aafe073f6c49e99ea9db935c2f257R5

PR #16497, opened 2017-02-22
update info.ini

tomsonpl:


tomsonpl:
@megawac  how long does it take for my files to be uploaded by a bot?

megawac:
20 or 30 minutes. In this case its not syncing because the update.json points at a `lib/` folder which doesnt exist

https://github.com/Syncano/syncano-client-js/tree/v.0.13.2-0

PR #16543, opened 2017-02-24
Add lg-thumbnail.js

Seig-Physics:


files/lg-thumbnail.js/update.json
```diff
@@ -0,0 +1,10 @@
+{
+  "packageManager": "github",
+  "name": "lg-thumbnail.js",
+  "repo": "sachinchoolur/lg-thumbnail.js",
+  "files": {
```
megawac: I think you need to do `basePath: 'dist/'` as well

files/lg-thumbnail.js/lg-thumbnail.js
```diff
@@ -0,0 +1,491 @@
+/**!
```
megawac: wrong folder, place under a version

PR #16658, opened 2017-03-01
Delete frinostyle.min.css

arbati6:


megawac:
why are we deleteing files. Regardless can we combine these 4 prs into 1

PR #16786, opened 2017-03-07
New "af-awesome" JS lib

ptraverse-aiq:


megawac:
Along with an info.ini file you should also either include an `update.json` or initial version of your project. See https://github.com/jsdelivr/jsdelivr/blob/master/CONTRIBUTING.md

PR #16845, opened 2017-03-11
Adding angular version 1.4.4

devendrant:


devendrant:
I am not getting why its failed. I haven't modified any file. Its been uploaded as it is, please review again.

PR #16888, opened 2017-03-13
Please add ecmwf-static-charts

ecmwf-sdo:
Many thanks


ecmwf-sdo:
Tag 1.0.0 added. Thanks

PR #17011, opened 2017-03-19
Add es6-tween

dalisoft:
* Adding es6-tween library. Thanks to @jsdelivr and contributors


dalisoft:
Please add this library faster (as previusly my projects)

PR #17029, opened 2017-03-20
Update domtastic dist files

webpro:
I've moved back to a single repo for this project. Are existing tags still picked up, or only the latest one?


megawac:
they should get picked up for major revisions.

PR #17144, opened 2017-03-22
Add learningjs

ImsirovicAjdin:


files/learningjs/files/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Ajdin Imsirovic"
```
megawac: place in `files/learningjs/info.ini`

PR #17163, opened 2017-03-23
Adding AutobahnJS

oddjobz:
Added a single file (autobahn.min.js). The current S3 bucket has just been pulled by the developers due to people linking live sites to it and racking up a bill. Typically this file is included with projects and hosted locally, but inevitably some people want to link to a hosted location, and in some cases a centrally updated host is a better choice.


oddjobz:
Ok, this is the minified version as delivered by the developers .. I think it needs to retain the copyright information uncompressed to comply with licensing .. which I think maybe why the checker is unhappy.

oddjobz:
Ok, this is the minified version, fed through UglifyJS2, with the copyright header pasted on afterwards .. it seems to be happy with this ... :)

oddjobz:
Do I need to do anything else here? How long does approval usually take?

MartinKolarik:
@jimaek this looks good I think?

PR #17249, opened 2017-03-28
Add Rrose leaflet plugin

GPavlovic:


megawac:
Thanks for the submission, cool library!

PR #17255, opened 2017-03-28
emailservice

mboosh:


files/emailservice/update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "packageManager": "github",
+  "name": "emailservice",
+  "repo": "mboosh/emailservice",
+  "files": {
+    "include": ["*.js", "*.css"]
```
megawac: You might want to set the basepath to the path to `webapp`

files/emailservice/update.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packageManager": "github",
+  "name": "emailservice",
+  "repo": "mboosh/emailservice",
+  "files": {
+    "basePath": "webapp",
```
megawac: from a glance I think this should be `src/main/webapp/`

files/emailservice/info.ini
```diff
@@ -0,0 +1,5 @@
+author = "Tim Yelland"
+github = "https://github.com/mboosh/emailservice"
+homepage = "https://github.com/mboosh/emailservice"
+description = ""
+mainfile = "src/main/webapp/js/app.js"
```
megawac: I think this is just js/app.js

mboosh: Thank you megawac for all your help, appreciate it as a first time user of jdelivr.

PR #17261, opened 2017-03-29
NPM isn't auto-updating. Moving to GitHub.

GoToLoop:
* Latest 0.58 has failed to auto-update -> https://github.com/processing/p5.js/releases/tag/0.5.8
* Changing it to get from GitHub now.
* Bonus: "p5.dom.min.js" & "p5.sound.min.js" minified versions are available there.
* Expecting JSDelivr can grab files directly from basePath: "releases/download/**/" though.

### DIFF
--- a/files/p5.js/update.json
+++ b/files/p5.js/update.json
@@ -1,9 +1,11 @@
 {
-  "packageManager": "npm",
+  "packageManager": "github",
   "name": "p5",
   "repo": "processing/p5.js",
   "files": {
-    "basePath": "lib/",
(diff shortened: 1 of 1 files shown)


megawac:
Not sure if the `releases/download/**/` will work but lets give it a shot. Otherwise might need to change the include paths to have the `**/*.js` instead

GoToLoop:
Thx @megawac! Well, we'll find that out once next 0.5.9 version comes out.  ;-)

PR #17345, opened 2017-04-01
CDN config files for WeUI.js

jloiola:


megawac:
Thanks for the submission! sorry for the delay merging

PR #17412, opened 2017-04-04
Delete webfonts.min.css

floydpink:


megawac:
We typically leave files, even if they're outdated on the cdn in case a user is relying on them. Just publish a new version and forget that this file exists

PR #17451, opened 2017-04-07
create update.json for MathJax

pkra:


megawac:
Awesome thanks, osrry for delay

SCHOLAR-Graeme:
@megawac Are you able to look at the MathJax folder:
https://github.com/jsdelivr/jsdelivr/tree/master/files/mathjax 
It's currently stuck on "Fetching latest commit…" and it would be good to get an update before the 30th of April before MathJax switch off their CDN.

megawac:
Hmm, @as-com any ideas? It looks good to me

as-com:
@megawac I don't see an images folder in the MathJax repository. Perhaps that's the issue?

megawac:
Hmm, good catch. I removed that line but still no dice it seems.

as-com:
Hmm...I'm just going to manually add the files. Gigantic pull request coming up!

as-com:
@megawac Do you think the problem is that MathJax simply has too many files?

megawac:
No idea, but there are some repos that have some 500+ files per version. How many mathjax is at?

megawac:
oh, uh jesus. what are all those files?

as-com:
@megawac Most of the files are the image fonts. MathJax uses images for older browsers.

pkra:
Sorry, I should have closed this after emailing with Dmitriy.

FWIW, the image fonts can be excluded if that helps. A quick fix might be to switch to npm as provider (we exclude them from the NPM package already).

PR #17556, opened 2017-04-13
Update project photon-themes to v0.1.3

jsdelivrbot:
libgrabber updated project photon-themes to v0.1.3.

@photonsh

### Metadata information

```json
{
  "metadata": {
    "packageManager": "github",
    "name": "photon-themes",
    "repo": "photonsh/themes",
    "files": {
      "include": ["**/*"],
      "exclude": []
    },
    "path": "/home/libgrabber/jsdelivr/files/photon-themes",
    "localVersions": ["0.1.0", "0.1.1", "0.1.2"],
    "remoteVersions": ["v0.1.0", "v0.1.1", "v0.1.2", "v0.1.3"],
    "branchVersions": []
  },
  "metadataPath": "/home/libgrabber/jsdelivr/files/photon-themes/update.json",
  "versions": ["v0.1.3"],
  "updateRemote": true,
  "version": "v0.1.3",
  "updated": true,
  "updatePath": "/home/libgrabber/jsdelivr/files/photon-themes/0.1.3",
  "branch": "photon-themes/v0.1.3"
}
```

PR #17601, opened 2017-04-15
Update project photon-themes to v0.1.3

felixsanz:
I think manual intervention is needed to fix https://github.com/jsdelivr/jsdelivr/pull/17556. So i've created this PR to save you some time/effort. Please tell me if something is wrong so i can fix it.

Thanks!

TL;DR: This should be merged instead of the automatic update from PR #17556. (I made a mistake and the bot PR is the bad one)


megawac:
Thanks for working this out

PR #17715, opened 2017-04-20
Add all versions of MathJax

as-com:
Since MathJax's CDN is shutting down, it would be nice to have a complete copy of all of the versions of MathJax, so jsDelivr could be a drop-in replacement for it.

I have added a copy of the important files from MathJax's Git repository, based on the files already hosted on cdnjs. I have also replaced the "2.0" and "2.1" folders with "2.0.0" and "2.1.0," respectively. This should not break anything, since the CDN should automatically alias requests from "2.1" to "2.1.0."

@jsdelivrbot will not like this pull request...


megawac:
jesus

megawac:
:see_no_evil: :speak_no_evil: :hear_no_evil:

PR #17738, opened 2017-04-21
all files added

subhransusekharswar:


megawac:
Hi @subhransusekharswar, if you're interested can you do another take on this. Version specific files should exist under

`/files/fusionchartjs/<version>/<file>`

Thanks!

PR #17762, opened 2017-04-23
add scrollmonitor to jsdelivr

misner:


misner:
I don't understand why the PR is rejected

megawac:
The bot didn't like your homepage. Typically you set it to the same as the github repo

PR #17870, opened 2017-04-29
Patch 1 of jquery.random-fade-in

sutara79:


sutara79:
I will close this PR because both "info.ini" and "update.json" are not in the same PR.
I'm sorry.

PR #17896, opened 2017-05-01
create line.js

futurecodesmith:


futurecodesmith:
I don't want auto update.  I just want my line.js file on CDN.

megawac:
Thats okay, if you decide to create tags later it will get picked up because you created the optional update.json

PR #17977, opened 2017-05-05
Create new datepicker plugin

mridah:


megawac:
Any reason you're adding so many versions of a project that doesnt have any documentation

mridah:
Hi Graeme,
Initially, we'll be using the plugin on our own website. And since the
developers were the same so knew how to use it. We've already started
writing the documentation and it should be finished soon :)

Also we added some new features and hence the new version.

PR #18000, opened 2017-05-10
Create update.json

MicroDreamIT:
requesting package to publish in jsdelivr


jimaek:
We no longer accept PRs. Read here how our new system works https://github.com/jsdelivr/jsdelivr#usage

PR #18054, opened 2017-12-29
Minor fix on string 289

kant:


jimaek:
Since they both point to the same place I think it makes sense to keep them 1 link? Otherwise people might think its 2 different endpoints which is not true

PR #18071, opened 2018-05-11
DraftOrderConfirmation goes on CDN

alemens:
I would like to get into your amazing CDN this two files: DraftOrderInvoice.svg and DINCondensed-Bold_2b which is a font. Thanks in advance


jimaek:
You don't need to do this. Just read our documentation, it explains how to use jsDelivr with github and npm

alemens:
Wow now I see my files! You are an amazing service. Thank you very much! I love you!

PR #18282, opened 2021-02-17
legal text

jimaek:
Since we are becoming more mainstream a lawyer suggested we make sure we are not liable for any problems. 
Its just a standard legal text that all companies have on their sites.
jsDelivr will always take all possible steps to prevent problems with accessing our service.

### DIFF
--- a/Acceptable Use Policy - jsdelivr.net.md
+++ b/Acceptable Use Policy - jsdelivr.net.md
@@ -1,6 +1,6 @@
 # Acceptable Use Policy
 
-*Effective date: August 15, 2020*
+*Effective date: February 17, 2021*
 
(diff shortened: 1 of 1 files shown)


jimaek:
I guess we also need to redeploy the main site after this is merged to update the text there?

MartinKolarik:
This seems more suited to be in ToS document which we don't have yet and I was thinking about recently because of jsdelivr/jsdelivr-backend#191

jimaek:
I know, but since we dont have a TOS, here should be fine as well. It can be anywhere really.

MartinKolarik:
Ok for now let's keep it simple. The website updates automatically from this doc.

MartinKolarik (approved):

PR #18296, opened 2021-04-24
fix(docs): corrects typo in project README

plan-do-break-fix:
Scope of changes is restricted to docs only.


jimaek:
Thank you for the fix!

PR #18319, opened 2021-08-16
Readme update

jimaek:
I merged the website documentation and our readme as well as added some small fixes. 

I am thinking of making this readme the official documentation and completely removing the "features" page from the website. Instead we could add small summaries of each important part of the documentation on the redesigned homepage.

@MartinKolarik What do you think?


MartinKolarik:
👍 to readme changes, probably 👎 to the website part but we can discuss that later.

PR #18467, opened 2023-02-10
policy updates

### LINKED ISSUE #18461: Update privacy policies
Both are a bit outdated in terms of the used third-party providers and services.

jimaek:
- No more Google analytics
- Not sure why even had CDNs in the .com policy since that part is covered by the .net policy
- Added Gcore

fixes #18461

### DIFF
--- a/Privacy Policy - jsdelivr.com.md
+++ b/Privacy Policy - jsdelivr.com.md
@@ -1,6 +1,6 @@
 # Privacy Policy - *.jsdelivr.com
 
-*Effective date: September 14, 2021*
+*Effective date: April 16, 2023*
 
 Prospect One ("us", "we", or "our") operates the jsdelivr.com website and its subdomains (the "Service").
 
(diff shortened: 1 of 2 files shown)


MartinKolarik:
The website uses (will use) the cdn too, so it's relevant.

jimaek:
>We use third-party Service Providers to serve all of our traffic under the domain cdn.jsdelivr.net.

We clearly define right now that there 2 policies and the user needs to select the domain. 

>The website uses (will use) the cdn too, so it's relevant.

Which CDN it will use? We can add that specific one to the .com policy

Privacy Policy - jsdelivr.com.md
```diff
@@ -190,42 +190,6 @@ to provide the Service on our behalf, to perform Service-related services or to
 These third parties have access to your Personal Data only to perform these tasks on our behalf
 and are obligated not to disclose or use it for any other purpose.
 
-### Analytics
-
-We may use third-party Service Providers to monitor and analyze the use of our Service.
-
-#### Google Analytics
-
-Google Analytics is a web analytics service offered by Google that tracks and reports website traffic.
```
MartinKolarik, line 199: This may need to be replaced by Plausible, or do we self host that?

jimaek: Self host

MartinKolarik:
Yes, the domain in the text was wrong, but I think we had both providers there to avoid the need to change the policy as we switch them.

> Which CDN it will use? We can add that specific one to the .com policy

I'm guessing Gcore, certainly not Bunny.

MartinKolarik:
Merged with updated CDN providers for www to avoid having to update the text every time we switch them. Ping me in DM if you think we should keep just Gcore there.

Issue #18475, opened 2023-04-15
mime question


54ayao:
Refused to execute script from 'https://cdn.jsdelivr.net/npm/typed.js' because its MIME type ('application/node') is not executable, and strict MIME type checking is enabled.

mime There's a problem.

54ayao:
> Refused to execute script from 'https://cdn.jsdelivr.net/npm/typed.js' because its MIME type ('application/node') is not executable, and strict MIME type checking is enabled.
> 
> mime There's a problem.

54ayao:
I do not know why this is the case, I think it should not be my problem, direct access to the JS file, the browser directly downloaded, I feel that part of the NPM may not be normal, hope to solve it as soon as possible, thank you

MartinKolarik:
You're loading a node.js file. The correct file for browsers is https://cdn.jsdelivr.net/npm/typed.js@2.0.15/dist/typed.umd.js

PR #18476, opened 2023-04-16
Better issue template

PencilNavigator:
i've gone ahead and remake the issue template in yaml format.

The yaml format issue template looks more user-friendly, and it is also easier for both users and maintainer to view. Also, The YAML fomat of issue forces user to submit it using the template, so we can avoid "invalid template issues" like https://github.com/jsdelivr/jsdelivr/issues/18475.

 you can try making an issue with the newest template at https://github.com/PencilNavigator/jsdelivr/issues/new/choose
 
 I also recommend turning on Github Discussions and remove other.md/other.yml template since most times things reported using that template is not really considered a "issue".


MartinKolarik:
Thanks, I made a few changes and will merge, including Other, for now. We'll discuss using the discussions feature later.

PR #18635, opened 2025-03-30
Fix incorrect spelling of Cloudflare

LitoMore:
See https://www.cloudflare.com/trademark/. 

"Cloudflare" is a trademark. We'd better use the correct spelling.


An0mol:
How would I do it?

